PR #36, opened 2026-04-17
fix: auto-discover sessions from multiple CLAUDE_CONFIG_DIR paths

### LINKED ISSUE #12: Does not detect claude instances if claude's directory is a custom path
## Context

I use two separate Claude Code configurations:
- `~/.claude` for my personal plan
- `~/.claude-pro` for my enterprise plan

I switch between them using aliases like:
```bash
alias claude-pro='CLAUDE_CONFIG_DIR=~/.claude-pro claude'
```

## Problem

`abtop` seams to hardcode the sessions and projects directories to `~/.claude/sessions` and `~/.claude/projects`.

When I launch Claude from my enterprise configuration (`CLAUDE_CONFIG_DIR=~/.claude-pro`), `abtop` does not detect the running sessions because it keeps looking in `~/.claude/` instead of `~/.claude-pro/`.

## Expected behavior

`abtop` should respect the `CLAUDE_CONFIG_DIR` environment variable — which is the official Claude Code mechanism for overriding the config directory — and use it to resolve the sessions and projects paths at startup.

For example, if `CLAUDE_CONFIG_DIR=~/.claude-pro` is set, `abtop` should look in `~/.claude-pro/sessions` and `~/.claude-pro/projects`.

## Suggested fix

In `src/collector/claude.rs`, the `ClaudeCollector::new()` function currently does:

```rust
let home = dirs::home_dir().unwrap_or_default();
Self {
    sessions_dir: home.join(".claude").join("sessions"),
    projects_dir: home.join(".claude").join("projects"),
    ...
}
```

It could be updated to check for `CLAUDE_CONFIG_DIR` first:

```rust
let base = std::env::var("CLAUDE_CONFIG_DIR")
    .map(PathBuf::from)
    .unwrap_or_else(|_| dirs::home_dir().unwrap_or_default().join(".claude"));
Self {
    sessions_dir: base.join("sessions"),
    projects_dir: base.join("projects"),
    ...
}
```

This would make `abtop` consistent with how Claude Code itself handles multi-account setups.

tbouquet:
## Summary

Fixes #12 - abtop now automatically discovers sessions from all active Claude Code config directories, not just one.

## Problem

Users running multiple Claude Code configurations (e.g. `~/.claude` for personal, `~/.claude-pro` for enterprise via `CLAUDE_CONFIG_DIR` aliases) only see sessions from whichever single directory abtop happens to resolve at startup. The current `CLAUDE_CONFIG_DIR` support (added in v0.2.6) requires abtop itself to have the variable set, which doesn't work when multiple instances use different values.

## Solution

Instead of reading `CLAUDE_CONFIG_DIR` from abtop's own environment, discover it dynamically from each running Claude process:

1. On each tick, read `/proc/<pid>/environ` for every detected Claude process
2. Extract `CLAUDE_CONFIG_DIR` from each process's environment
3. Scan sessions, transcripts, and rate limit files from **all** unique config directories

Always includes `~/.claude` (default) plus any `CLAUDE_CONFIG_DIR` set in abtop's own env as baseline.

## Changes

- `ClaudeCollector` now maintains a `Vec<ConfigDir>` refreshed each tick
- `find_transcript` searches across all discovered project directories
- `rate_limit::read_rate_limits` checks all known config dirs
- New `read_env_var_from_proc()` reads `/proc/<pid>/environ` (Linux-only, with non-Linux stub)

## Testing

- `cargo clippy -- -D warnings` - clean
- `cargo test` - 31/31 pass
- `cargo build --release` - compiles, binary functional
- Runtime: `abtop --once` correctly shows all 4 active sessions
- `/proc/<pid>/environ` reading verified on live system (68 env vars readable)

## Limitations

- `/proc` reading is Linux-only. On macOS, discovery falls back to `CLAUDE_CONFIG_DIR` env var + default `~/.claude` (same as current behavior). A future enhancement could use `sysctl kern.procargs2` on macOS.
- Rate limits require `abtop --setup` to have been run in each config directory separately (each Claude instance needs its own statusline hook).

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -129,7 +129,8 @@ impl App {
         // Poll rate limits: first tick immediately, then every 5 ticks ≈ 10s
         if self.rate_limits.is_empty() || self.rate_limit_counter >= 5 {
             self.rate_limit_counter = 0;
-            self.rate_limits = read_rate_limits();
+            let extra_dirs = self.collector.all_config_dirs();
+            self.rate_limits = read_rate_limits(&extra_dirs);
             // Merge live rate limits from agent collectors (e.g. Codex JSONL parsing)
             self.rate_limits.extend(self.collector.agent_rate_limits());
         } else {
--- a/src/collector/claude.rs
+++ b/src/collector/claude.rs
@@ -72,7 +72,12 @@ impl ClaudeCollector {
     }
 
     fn collect_sessions(&mut self, shared: &super::SharedProcessData) -> Vec<AgentSession> {
-        self.refresh_config_dirs(&shared.process_info);
+        // Refresh config dirs on slow ticks only (every ~10s) or on first run.
+        // Scanning /proc/<pid>/environ for every Claude process is expensive
+        // to do every 2s; config dirs change rarely.
+        if shared.slow_tick || self.config_dirs.is_empty() {
+            self.refresh_config_dirs(&shared.process_info);
+        }
 
         // Collect all session file paths first to avoid borrowing self
         // immutably (config_dirs) and mutably (load_session) at the same time.
@@ -93,9 +98,12 @@ impl ClaudeCollector {
         }
 
         let mut sessions = Vec::new();
+        let mut seen_ids = std::collections::HashSet::new();
         for path in &session_paths {
             if let Some(session) = self.load_session(path, &shared.process_info, &shared.children_map, &shared.ports) {
-                sessions.push(session);
+                if seen_ids.insert(session.session_id.clone()) {
+                    sessions.push(session);
+                }
             }
         }
 
@@ -519,6 +527,12 @@ impl ClaudeCollector {
 impl super::AgentCollector for ClaudeCollector {
     fn collect(&mut self, shared: &super::SharedProcessData) -> Vec<AgentSession> {
         self.collect_sessions(shared)
+    }
+
+    fn discovered_config_dirs(&self) -> Vec<PathBuf> {
+        self.config_dirs.iter()
+            .map(|c| c.sessions_dir.parent().unwrap_or(Path::new(".")).to_path_buf())
+            .collect()
     }
 }
 
--- a/src/collector/claude.rs
+++ b/src/collector/claude.rs
@@ -34,7 +34,8 @@ impl ClaudeCollector {
     /// /proc/<pid>/environ for each running Claude process.
     /// Always includes the default (~/.claude) and CLAUDE_CONFIG_DIR if set.
     fn refresh_config_dirs(&mut self, process_info: &HashMap<u32, process::ProcInfo>) {
-        let mut seen = std::collections::HashSet::new();
+        // BTreeSet for deterministic iteration order across runs.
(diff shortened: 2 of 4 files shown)


graykode (commented):
Thanks so much for tackling this! The `/proc/<pid>/environ` approach is a really nice way to solve the multi-alias workflow from #12 — much better than requiring abtop itself to know about every config dir ahead of time.

A few thoughts I had while reading through, mostly small:

**1. Iteration order of `config_dirs`**

`config_dirs` is built from a `HashSet`, so the order isn't stable across runs. I think this could subtly affect two spots:

- In `find_transcript`, a worktree-fallback match in one dir could shadow an exact encoded-cwd match in another dir, depending on order. Splitting it into two passes (try encoded-cwd across all dirs first, then the fallback scan across all dirs) would make this deterministic.
- `load_session`'s fallback picks `config_dirs.first()`, which ends up being effectively random.

Switching to `BTreeSet<PathBuf>` or a `Vec` + `.contains()` should cover both.

**2. Scope of `Closes #12`**

Session discovery is definitely fixed, but `rate_limit.rs` still only reads from abtop's own `CLAUDE_CONFIG_DIR` + `~/.claude`, so the QUOTA half of #12 (where @hmenzagh originally reported the quota column staying empty) isn't fully addressed yet. Totally fine either way — either threading the discovered config dirs through to `read_rate_limits`, or switching `Closes #12` to `Refs #12` and leaving quota for a follow-up works for me.

**3. macOS coverage note**

abtop ships for both Linux and macOS (aarch64/x86_64 each). On macOS the stub returns `None`, so discovery falls back to abtop's own env var — essentially the v0.2.6 behavior. The PR builds cleanly everywhere, which is great, but if the original reporter happens to be on macOS, they wouldn't see an improvement yet. Could you add a line to the Limitations section making this explicit? A macOS version via `sysctl kern.procargs2` would be a nice follow-up but definitely out of scope here.

**Nice-to-haves (not blockers)**
- Moving `refresh_config_dirs` to the slow poll tick would be consistent with how git stats / ports are handled.
- A small unit test for `read_env_var_from_proc` covering the NUL-separated parsing and the "var not set" case would be nice.

Again, really appreciate the work here — the direction is exactly right. Happy to merge once the first point is addressed; the rest can be follow-ups if you'd rather.

graykode:
Pushed a one-liner (7641289) to gate `parse_environ_var`'s `dead_code` suppression behind `cfg(not(target_os = "linux"))` so macOS `cargo clippy -- -D warnings` stays clean while Linux still catches real dead-code regressions. Verified locally: 35/35 tests pass, clippy clean on macOS. Merging now — thanks again for the great work on this one!

graykode (approved):
All review feedback has been addressed thoroughly. Deterministic iteration via BTreeSet, two-phase find_transcript, rate limit asymmetry fixed via discovered_config_dirs trait + all_config_dirs aggregation, slow-poll refresh, and unit tests for parse_environ_var including the no_partial_match edge case. The macOS dead_code fix has been pushed and verified (35/35 tests, clippy clean). LGTM.

PR #37, opened 2026-04-17
chore: reduce tokio features to rt+macros

tbouquet:
## Summary

- Reduce tokio features from `"full"` to `["rt", "macros"]`
- The codebase is fully synchronous (`fn main`, `std::thread::spawn`, `std::sync::mpsc`) with zero uses of `async`/`.await`/`tokio::` anywhere in `src/`
- `"full"` pulls in unused modules (net, fs, io-util, signal, sync::broadcast) and ~9 transitive crates that only increase compile time and attack surface

## Test plan

- [x] `cargo test` - all 25 tests pass
- [x] Verified: `grep -r "async\|tokio\|\.await" src/` returns zero matches

## Additional note

A `cargo-audit` CI job (`rustsec/audit-check@v2`) would also be valuable but requires the `workflow` scope to modify `.github/workflows/ci.yml`. Happy to add it in a follow-up if you prefer, or you can add it directly:

```yaml
  audit:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: rustsec/audit-check@v2
        with:
          token: ${{ secrets.GITHUB_TOKEN }}
```

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -257,7 +257,6 @@ Tracks child processes that have open ports. When a parent session dies but the 
 - **Rust** (2021 edition)
 - **ratatui** + **crossterm** for TUI
 - **serde** + **serde_json** for JSON/JSONL parsing
-- **tokio** — included but currently sync I/O; summary generation uses `std::thread::spawn`
 - **chrono** for timestamp formatting
 - **dirs** for home directory resolution
 - **Polling intervals** (staggered to avoid freezes):
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,7 +13,6 @@ dependencies = [
  "serde",
  "serde_json",
  "tempfile",
- "tokio",
 ]
 
 [[package]]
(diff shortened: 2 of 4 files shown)


graykode:
Thanks @tbouquet — nice catch and the analysis is spot-on.

One suggestion before we merge: since `grep -rn "tokio\|async\|\.await" src/` returns **zero** matches, `tokio` is effectively an unused dependency. Trimming features to `["rt", "macros"]` is an improvement, but the logical conclusion is to **drop the dependency entirely** rather than keep an unused crate around.

Could you:
1. Remove the `tokio = ...` line from `Cargo.toml` altogether
2. Update the `tokio — included but currently sync I/O` note in `CLAUDE.md` accordingly

Also, the `user_message` match-guard refactor in `src/collector/codex.rs` is unrelated to the tokio change — ideally that would be a separate PR (or at least mentioned in the description), but it's a small stylistic win so no blocker.

`cargo-audit` CI addition sounds great as a follow-up PR, thanks for offering.

graykode:
@tbouquet One more thing — the change in `src/collector/codex.rs` looks unrelated to this PR's scope (reducing tokio features). Could you revert that file so this PR stays focused on the `Cargo.toml` change? Happy to review the match-guard refactor in a separate PR.

graykode (approved):

Issue #3, opened 2026-03-31
Issues with workspaces


yawn:
Great tool, especially now in the times of the usage apocalypse!

When using Claude Code with `--workspace` the tool fails to detect metadata such as model, context, tokens or turns (memory usage is detected). To reproduce just start any session with `--workspace` and see the difference.

yawn:
Addendum: it also show no git metadata for worktrees.

Issue #2, opened 2026-03-30
Fix: Claude sessions not detected when process shows as 'claude' without path


graykode:
## Bug

`abtop --once` shows 0 sessions even though `~/.claude/sessions/` contains active session files.

## Root Cause

The process matching at `src/collector/claude.rs:69` required both `/claude` (with path prefix) and `--session-id` in the `ps` command string:

```rust
c.contains("/claude") && c.contains("--session-id")
```

Some installations (e.g. Homebrew on macOS) show the process as just `claude` without a full path and without `--session-id` flag.

## Fix

Changed matching to accept both formats:

```rust
c.starts_with("claude") || c.contains("/claude")
```

Fixed in 3a988a5.

Issue #1, opened 2026-03-30
demo mp4


graykode:
https://github.com/user-attachments/assets/d50b7c1b-ed81-44a2-903f-902d4de911ae

PR #39, opened 2026-04-18
feat: config page overlay and panel toggle (Esc, 1-5)

tbouquet:
## Summary

- **Esc** opens a config overlay to change theme and toggle panels
- **1-5** toggles individual panels directly without opening config:
  - `1` Context  `2` Quota  `3` Tokens  `4` Ports  `5` Sessions
- Hidden panels redistribute space to visible ones
- Config overlay captures input while open (Esc/q to close)
- Footer updated with new keybinding hints

## Demo

config-demo

Shows: panel toggle (2/3 keys), config overlay (Esc), theme cycling, panel toggle from config

## Design

- Pure UI change: no file I/O, no process execution, no new dependencies
- Panel visibility is stored as bools on `App` (session-scoped, not persisted)
- Layout recalculates dynamically when panels are hidden
- Projects panel always visible (provides essential git context)

## Test plan

- [x] `cargo test` - all 31 tests pass
- [x] Rebased on latest main (v0.2.12)
- [x] Security reviewed: no I/O, no Command, no fs ops in diff


graykode (commented):
Thanks for this! Really like the direction — an Esc overlay + number keys is a nice way to let users shape the layout, and the dynamic redistribution works well.

A few things I noticed while reading through:

**1. `q` in the config overlay quits abtop (small bug)**

The PR description says "Esc/q to close" the overlay, but in `main.rs` the `q` branch inside the `config_open` block calls `app.quit()` instead of `app.toggle_config()`. So pressing `q` with the overlay open kills abtop rather than dismissing the overlay. I think the fix is just routing `q` to `toggle_config()` while the overlay is open.

**2. Hiding all three toggleable mid panels also hides Projects**

The PR description notes "Projects panel always visible (provides essential git context)", but:

```rust
let any_mid = app.show_quota || app.show_tokens || app.show_ports;
let mid_h_ideal: u16 = if any_mid { 8 } else { 0 };
```

If a user toggles off Quota, Tokens, and Ports, `any_mid` becomes false, `mid_h` becomes 0, and the whole mid row — including Projects — disappears. I think `any_mid` should effectively be `true` whenever Projects is shown, or the variable can just be dropped so the mid row always renders at least Projects.

**3. Toggle key numbering vs. panel numbering in CLAUDE.md**

CLAUDE.md documents the panels as ¹context ²quota ³tokens ⁴projects ⁵ports ⁶sessions, but the toggle keys are 1=context, 2=quota, 3=tokens, 4=ports, 5=sessions (skipping Projects). Users glancing at the layout diagram may expect `4` to toggle Projects and `5` to toggle Ports. Two possible fixes: (a) keep the CLAUDE.md numbering and leave `4` unbound since Projects is always on, or (b) update the CLAUDE.md panel numbers to match this PR's scheme. Either works — just worth being consistent.

**4. `Esc` as an "open" key is a little unusual**

In most TUIs, `Esc` means close/cancel. Using it to both open and close the config overlay works, but a more conventional pairing would be `c` (or `?`) to open and `Esc` to close. Not blocking, just a small UX note — happy either way.

**Nice-to-haves (not blockers)**

- The mid constraint construction in `ui/mod.rs` builds a `Vec` of `Length(0)` dummies just to count them, then rebuilds as `Ratio(1, count)`. Could be simplified to counting directly:
  ```rust
  let count = 1 + app.show_quota as u32 + app.show_tokens as u32 + app.show_ports as u32;
  let mid_constraints: Vec<Constraint> = (0..count).map(|_| Constraint::Ratio(1, count)).collect();
  ```
- The `codex.rs` clippy fix (match guard for `user_message`) is also included in #36, so whichever lands second will need a small rebase — nothing to do here, just a heads up.
- A unit test for `toggle_panel` / `config_select_next` / `config_select_prev` would be easy and would lock in the bounds behavior (e.g. `config_selected` not overflowing `config_item_count`).
- `config_item_count` hardcodes `6`; deriving it from the item list (or a const) would keep it in sync if you add more options later.
- Persisting panel visibility across runs (e.g. `~/.config/abtop/config.toml`) would be a natural follow-up.

Thanks again — once #1 and #2 are addressed I think this is in great shape to merge. The rest are all discussable / follow-up material.

graykode (approved):
Thanks for the thorough follow-up! All review feedback addressed:

- `q` in the overlay now correctly routes to `toggle_config` (blocker #1)
- Mid row always renders since Projects is always visible (blocker #2)
- Config open key changed to `c`; `Esc` reserved for close/cancel (UX #4)
- CLAUDE.md panel numbering aligned with toggle keys: Projects unnumbered (always visible), Ports=4, Sessions=5 (UX #3)

Merge conflict with main (caused by PR #36 landing first and the prior redact_secrets hardening from #35) resolved by taking main's version, which is a superset of this PR's clippy fix. Verified locally: 35/35 tests pass, clippy clean. Merging now.

PR #40, opened 2026-04-18
feat: add OpenCode agent support

### LINKED ISSUE #38: Support of OpenCode
**Describe the feature**

Support of OpenCode which is a widely used OSS coding agent

**Use case**

See https://opencode.ai/

**Additional context**

From link above: "With over 140,000 GitHub stars, 850 contributors, and over 11,000 commits, OpenCode is used and trusted by over 6.5M developers every month."

tbouquet:
## Summary

Closes #38 - adds [OpenCode](https://opencode.ai/) as a supported agent alongside Claude Code and Codex CLI.

- New `OpenCodeCollector` discovers sessions from OpenCode's SQLite database (`~/.local/share/opencode/opencode.db`)
- Uses `sqlite3` CLI with `-readonly -json` flags - **no new Rust dependencies** (no rusqlite/libsqlite3)
- Process discovery, PID matching, status derivation, child collection, and kill verification follow existing Claude/Codex patterns
- Emerald `#OC` label in sessions panel
- Demo mode includes an OpenCode example session
- Graceful fallback when `sqlite3` binary or `opencode.db` are unavailable
- Documents that quota remains Claude + Codex only because OpenCode does not expose reliable account-level provider quota data

## Design decisions

| Decision | Rationale |
|----------|-----------|
| sqlite3 CLI over rusqlite crate | Avoids C dependency (libsqlite3-sys), keeps binary lean, follows existing pattern of shelling out to ps/lsof/git |
| `-readonly` flag | Safe concurrent reads with WAL mode, no locking |
| GROUP BY + separate model lookup | Single-pass token aggregation; model/provider are read separately from the latest assistant row |
| cwd-based PID matching | OpenCode does not expose a PID/session mapping; live PIDs are assigned at most once to avoid showing old DB rows as duplicate live sessions |
| no OpenCode quota row | OpenCode can use many providers and does not expose a stable account-level quota source; quota remains Claude + Codex only |

## Test plan

- [x] `cargo test` - 121 passed, 0 failed
- [x] CI `check` passes
- [x] CI `plan` passes
- [x] Verified against real OpenCode SQLite DB (v1.4.0) with 3 sessions, 55 messages
- [x] Verified graceful fallback when sqlite3 is absent (returns empty vec)
- [x] Demo mode renders OpenCode session with emerald label

## Notes

- `cargo fmt -- --check` currently reports pre-existing formatting differences in files outside the final patch scope (`claude.rs`, `codex.rs`, `demo.rs`, `ui/mcp.rs`, `ui/sessions.rs`). CI does not currently require fmt.


graykode (changes requested):
Thanks for another substantial PR, @tbouquet. OpenCode support is a great addition and the shell-out-to-sqlite3 approach is the right call for keeping the binary lean. A few blocking issues before merge:

### 🔴 Blocking

**1. JSON output parsing is fragile**

```rust
let arrays: Vec<&str> = stdout.trim().split("][").collect();
```

OpenCode auto-generates session `title` from user prompts, so a title containing `][` (or a directory path with `][`) will break parsing silently. Two clean options:

- **One query with LEFT JOIN**: fold the `model`/`provider` latest-row subqueries into the main SELECT as correlated subqueries in the column list. Then you only get one JSON array back.
- **Two separate `sqlite3` invocations**: cheaper to reason about, ~2x fork cost but it's on the slow tick.

Either way, avoid splitting concatenated JSON by a literal substring.

**2. `/proc/{pid}/cwd` is Linux-only — macOS is abtop's primary target**

```rust
if let Ok(cwd) = std::fs::read_link(format!("/proc/{}/cwd", pid)) {
```

On macOS this always returns `Err`, so PID→session matching falls through to the (weak) command-line substring fallback and finally the "single opencode process" heuristic. With multiple opencode instances on macOS, matching will be wrong.

Suggested fix: use `lsof -p <pid> -d cwd -Fn` as the macOS path. We already use `lsof` for port scans, so it's consistent. Gate the `/proc` path behind `#[cfg(target_os = "linux")]` and use `lsof` otherwise.

**3. Missing security parity with #35 (just merged)**

#35 added hardening to Claude and Codex collectors. OpenCode needs the same treatment:

- **Symlink skip on `db_path`**: if `~/.local/share/opencode/opencode.db` is a symlink, fail closed (skip). Matches the pattern in `claude.rs`.
- **Truncate DB string fields** (`title`, `directory`, `version`, `model`, `provider`) at sane char boundaries after deserialization, similar to `SessionFile::sanitize()` in `model/session.rs`.
- **Redact secrets in `initial_prompt`**: `ds.title` becomes `initial_prompt`, which is displayed in the TUI. Pipe it through `collector::redact_secrets` (now public in `collector/mod.rs`).
- **Sanitize output**: control chars / bidi overrides in DB-sourced strings should be stripped on display. `main::sanitize_output` already handles this for `--once`, but the TUI path should also apply it to OpenCode session fields.

**4. Rebase needed — Codex change reverts #37's clippy fix**

The diff in `src/collector/codex.rs` reverts `Some("user_message") if ... =>` back to nested form. main has it the other way (clippy 1.95 `collapsible_match`). Rebase onto main and take main's version.

### 🟡 Nice to have (non-blocking)

- **`time_updated` unit**: please add a one-line comment confirming it's ms (vs seconds). Current `saturating_sub` assumes ms.
- **Two separate `LIMIT 20` subqueries**: if `time_updated` changes between them, the id sets can diverge and some sessions lose their `model`/`provider`. LEFT JOIN (per #1 above) also fixes this.
- **`sqlite3_available` cached forever**: consider revalidating every few minutes so users who install `sqlite3` mid-session don't have to restart abtop.

### 🟢 Things that are good

- No new Rust dependencies, consistent with existing shell-out patterns ✓
- `-readonly` + WAL-safe reads ✓
- `visited` HashSet on child walker (learned from #35 pattern) ✓
- Trait abstraction followed cleanly ✓
- Avoids the `Box::leak` path in the sessions UI by adding an explicit `"opencode"` arm ✓

Let me know if you want to discuss the design choices on #1 or #2 before implementing — happy to iterate.

graykode (changes requested):
picked this back up after consolidating #100 into this branch — the review-feedback commit is in, plus a merge commit for the rebase against current main. security still clean (sqlite3 args are constants, lsof pid is u32, db_path symlink fail-closed, secrets redacted on title). 116/116 tests pass on the merged tip.

a few correctness issues remain in opencode.rs that didn't surface in the earlier round:

1. **`match_pid_to_session` last-resort over-matches every session.** when exactly one opencode process is running but its cwd doesn't match a session's `directory`, the `if pid_commands.len() == 1` branch returns that PID for *every* call. so if the user has N rows in the DB and one live opencode process working in a different dir, all N rows render as alive (Thinking/Waiting) reusing the same PID. the fallback should either be dropped entirely or only kick in when there's exactly one DB session being matched (not per-row).

2. **`cmd.contains(session_dir)` is unguarded against empty/short strings.** in the same function: an empty `session_dir` makes `cmd.contains("")` always true, and a `directory` of `/` matches every absolute path. add `if !session_dir.is_empty() && session_dir.len() > 1` (or similar) before the substring fallback.

3. **doc comment is out of sync with the code.** the `OpenCodeCollector` docstring says "Queries run on slow ticks only (every ~10s via MultiCollector gating) to avoid forking a sqlite3 process every 2s", but `collect()` never checks `shared.slow_tick` — sqlite3 forks every tick (2s). either gate the queries on `shared.slow_tick` and cache the previous result, or fix the comment to match reality.

4. **"recently finished < 5 min" branch is dead code.** the `if !pid_alive && age_ms > 300_000 { continue; }` block keeps DB rows whose PID isn't alive, sets `status = SessionStatus::Done`, and tags `current_tasks = vec!["finished"]`. but `mod.rs:273` does `all.retain(|s| !matches!(s.status, SessionStatus::Done))` — those rows never reach the UI. either drop the dead branch, or follow the codex pattern (pid==0 sentinel + status that isn't Done) if showing recently-finished rows is intended.

nits: `since_update_secs < 30 → Thinking` is a coarse heuristic vs codex's `model_generating`, but acceptable for a TUI dashboard. the children-walk logic duplicates codex.rs verbatim and could share a helper, not blocking.

requesting changes for #1 specifically (user-visible bug with multiple DB sessions). #2-#4 are easy follow-ups in the same push. happy to take it from here if you'd rather not — let me know.

graykode:
Pushed `753d42f` to tighten the OpenCode UI/collector behavior after the quota discussion.

Changes:
- Documented in `AGENTS.md` that quota is intentionally Claude + Codex only; OpenCode contributes session/token/project/port data, not quota data.
- Allowed `x` kill verification to recognize `opencode` processes instead of only `claude`/`codex`.
- Prevented one live OpenCode PID from being assigned repeatedly to multiple DB session rows from the same cwd, which would render old sessions as live duplicates.
- Added regression tests for the OpenCode kill whitelist and one-PID-per-session-row matching.

Verification:
- `cargo test` -> 121 passed, 0 failed
- `cargo clippy -- -D warnings` -> passed
- `cargo fmt -- --check` still reports pre-existing formatting differences in files outside this patch scope (`claude.rs`, `codex.rs`, `demo.rs`, `ui/mcp.rs`, `ui/sessions.rs`).

graykode (approved):
Reviewed latest head 753d42f after the OpenCode quota/UI follow-up. The previous blocking items are addressed, CI is green, and local verification passed with cargo test and clippy.

PR #42, opened 2026-04-19
fix: paint theme background for light-terminal readability

yrashk:
## Summary
- On light-background terminals, abtop's UI was hard to read because the themes only set foreground colors and relied on the terminal's bg being dark.
- Added a `main_bg` field to `Theme` with a sensible dark value for each preset (btop #191919, dracula #282a36, catppuccin #1e1e2e, tokyo-night #1a1b26, gruvbox #282828, nord #2e3440, high-contrast #000, and matching dark bgs for the colorblind-safe variants).
- `ui::draw` now paints the entire frame with `theme.main_bg` before rendering panels, so the UI looks the same regardless of the underlying terminal theme.

## Test plan
- [x] `cargo build`
- [x] `cargo clippy` (clean)
- [ ] Run `cargo run` in a light-background terminal (e.g. default macOS Terminal "Basic") and confirm the UI is readable
- [ ] Cycle themes with `t` and confirm each theme renders with its dark background in both light and dark terminals


graykode (approved):
Thanks for this — real issue on light terminals. Pushed a one-liner that fills the config overlay's interior with `theme.main_bg` so it matches the new frame background (without it, `Clear` leaves the popup interior at terminal-default bg, so light-terminal users would see a bright rectangle in the middle of the dark-themed app). Verified locally: clippy clean, 35/35 tests pass. Merging now.

Follow-up idea (separate PR welcome): `main_bg: Option<Color>` or a "native" theme preset with `Color::Reset` so users who intentionally want their terminal's background (e.g. for accessibility) can opt out.

PR #49, opened 2026-04-19
feat: tool use timeline view (L key toggle)

tbouquet:
## Summary

Press **L** to toggle a tool use timeline in the session detail panel - LangSmith-style trace view, but in the terminal.

## Demo

timeline-demo

Shows: normal view -> L toggles timeline with colored duration bars -> navigate between sessions -> L toggles back

## Features

- Parses **all** tool_use entries from Claude transcripts (previously only the last one)
- Computes duration from assistant turn timestamp to next user turn timestamp
- Duration bars scaled relative to the longest call, marked with **\***
- Each tool type gets a distinct color:
  - **Read** = blue, **Edit** = yellow, **Bash** = green
  - **Write** = cyan, **Grep** = purple, **Agent** = coral
- Header shows total call count and cumulative duration
- Capped at 500 tool calls per session to prevent OOM
- Incremental parsing: new tool_calls merged via transcript cache delta

## Data model

- New `ToolCall` struct: `name`, `arg`, `duration_ms`
- `AgentSession.tool_calls: Vec<ToolCall>`
- Duration computed by comparing `assistant` turn timestamp with next `user` turn timestamp

> "Horizontal span view of tool calls per session with duration bars. LangSmith in the terminal. No TUI tool does this."

## Test plan

- [x] `cargo test` - 31/31 pass
- [x] Demo mode: webshop session has 16 realistic tool calls with duration bars
- [x] Real transcripts: tested against active Claude Code session with 279 tool_use entries


graykode (changes requested):
nice feature. security is clean (no new deps, network calls, or shell); local build/clippy/tests all green (35/35). a few things before I approve:

- `src/app.rs:75-76` — `timeline_scroll` is stored but nothing ever increments it. with `l` toggled and any session that has more tool calls than the viewport rows (likely for most real sessions, since the cap is 500), everything past the first screen is unreachable. either wire PgUp/PgDn/j/k-in-timeline-mode, or drop the field.
- `src/collector/claude.rs:209` — `prev.last_assistant_ts_ms = delta.last_assistant_ts_ms` overwrites unconditionally. if a transcript read boundary lands after an assistant turn but before its tool_result `user` line, the next parse starts with a fresh `last_assistant_ts_ms = 0` and the duration is silently lost (tool_call stays at 0ms). suggest only overwriting when delta actually saw a new assistant entry, or carrying `prev.last_assistant_ts_ms` into `parse_transcript`.
- `src/ui/sessions.rs:430-551` — the else-if refactor didn't re-indent the wrapped children/subagents block, so it sits one level too shallow inside `if let Some(lower) = lower_area`. minor but jumps out when reading.
- scope creep: `src/app.rs:72-74` strips the doc comments on `tree_view`/`filter_text`/`filter_active`, and `src/collector/claude.rs:730` collapses the `cache_creation` rationale (#54) into a terse one-liner. those explain the "why" and shouldn't vanish with a feature PR — please restore.
- `src/ui/sessions.rs:632-645` — `tool_color` reuses semantically-unrelated theme fields (`session_id` for Read, `cpu_box` for Write, `selected_fg` for Skill). on some themes `selected_fg` will collide with the selected-row foreground and `cpu_box` with actual box borders. a small dedicated palette (or reusing a few `proc_grad` stops) would be safer across themes.

happy to re-review once scroll is wired (or removed) and the dropped comments come back.

graykode (approved):

PR #50, opened 2026-04-19
feat: finer agent state indicator (Thinking/Executing/Waiting/RateLimited)

tbouquet:
## Summary

Replaces the coarse 3-state `SessionStatus` (Working/Waiting/Done) with a 5-state enum providing much finer visibility into what an agent is doing.

## Demo

agent-state-demo

## States

| State | Icon | When |
|-------|------|------|
| **Thinking** | `● Think` | Agent is generating (recent activity, no active tool) |
| **Executing** | `● Exec` | Running a tool (current_task is non-empty) |
| **Waiting** | `◌ Wait` | Idle, waiting for user input |
| **RateLimited** | `⏳ Rate` | Promoted from Waiting when any rate limit > 90% |
| **Done** | `✓ Done` | Session finished |

## Implementation

- `SessionStatus::Working` replaced by `Thinking` + `Executing` based on `current_task` emptiness
- `RateLimited` promoted from `Waiting` by App post-collection when `rate_limits > 90%`
- `is_active()` helper replaces all `matches!(status, Working)` checks
- Updated: Claude collector, Codex collector, UI sessions, header, context, snapshot mode, demo

## Test plan

- [x] No `SessionStatus::Working` references remain
- [x] Demo: webshop=Executing, ml-pipeline=Waiting, api-server=Executing, data-viz=Thinking


graykode (changes requested):
security clean (no new network/shell/deps). holding off approval — two correctness items, plus a scope question.

scope: the diff also adds a whole **file access audit** feature (`FileOp`/`FileAccess` in `src/model/session.rs`, `App.show_file_audit` + `f` keybinding in `src/main.rs:161`, ~38-line `draw_file_audit` renderer in `src/ui/sessions.rs:659`, tool_use→FileAccess parsing in `src/collector/claude.rs:1158-1180`, demo data). PR title/body only mention the state enum. would suggest splitting it into a second PR — both features are meaty enough to deserve their own review threads, and bisecting is easier later.

correctness:

- `src/app.rs:227-235`: RateLimited promotion checks `self.rate_limits.iter().any(...)` across all sources, then promotes every Waiting session regardless of agent type. since `RateLimitInfo.source` is "claude"|"codex" and `AgentSession.agent_cli` matches, a Codex session ends up RateLimited when only Claude is over 90%. needs a per-cli filter — something like `rl.source == s.agent_cli` inside the closure.
- `src/collector/claude.rs:479` (and `codex.rs:184`): the new stale branch (`has_active_descendant || has_tool → Executing`) classifies as Executing forever for sessions whose final assistant turn ended with a tool_use and never came back (interrupted/crashed). pre-PR behavior was Waiting in that case. niche but a regression — gating the `has_tool` arm on a freshness window would close it.

other:

- no unit tests for the new logic (RateLimited promotion, Thinking/Executing split, file_accesses parsing). the cross-agent bug above would be a one-liner test against the existing fixture pattern in `claude.rs`.
- "Thinking" status overlaps semantically with the live Think row from #67 (`thinking_since_ms`) — same word, different concepts. `Generating` or `Active` would read cleaner.
- `assets/agent-state-demo.gif` is 851KB committed to history. fine if it's the repo convention, but worth checking with @graykode whether release-asset / external hosting is preferred.

happy to revisit once scope is settled and the rate-limit filter lands.

tbouquet:
Thanks for the careful read. Pushed `a288769` (rebased onto the new main with PR #70 in) addressing the two correctness items + tests:

**1. Per-CLI rate limit promotion** (`src/app.rs`): extracted into `promote_waiting_to_rate_limited(sessions, rate_limits)` which now filters `rl.source == s.agent_cli`, so Claude saturation no longer freezes Codex and vice versa. Covered by three unit tests (per-cli filter, below-threshold no-op, non-Waiting skip).

**2. Stale-transcript `has_tool` guard** (`src/collector/claude.rs` + `codex.rs`): the stale branch no longer accepts `has_tool` alone. Only `has_active_descendant` promotes to Executing, so a crashed/interrupted session whose last turn ended on a tool_use falls back through Thinking (CPU) or Waiting instead of being stuck Executing. Commented the reasoning inline.

Remaining items, let me know how you'd like to proceed:

- **Scope split**: happy to carve the file-audit feature into its own PR if you prefer. It's cleanly separable (`FileAccess`/`FileOp` in model, `App.show_file_audit` + `F` binding, `draw_file_audit` renderer, tool_use parsing in `claude.rs`). Would push this one with just the state-enum work, and open a follow-up PR for file audit. Your call.
- **`Thinking` naming overlap with #67**: agree the collision is real. Open to `Generating` or `Active` - I'd lean `Generating` since `Active` doubles up with `is_active()`. Will rename in the scope-split PR if you want to wait, or do it here.
- **851KB GIF**: will check with you on the repo convention before touching. If release-asset / external hosting is the preferred pattern I'll swap it over.

graykode:
Pushed `4f7c30b` directly — small follow-up I noticed while looking at the file-audit code:

`Vec::truncate(MAX)` keeps the FIRST N elements, so the cache-extend at `claude.rs:392` and the per-push guard in `parse_transcript` were both keeping the *oldest* 1000 file accesses and silently dropping every new one once a long session crossed the cap. The UI's `.iter().rev().take(...)` would then look "frozen" because the latest entries never made it into the cache.

Switched both callsites to drain-from-front (sliding window), and lifted the parse-time cap to a single end-of-parse drain so peak memory during parse stays bounded by the file size. Added `test_parse_transcript_file_accesses_sliding_window` covering `MAX_FILE_ACCESSES + 5`: oldest 5 drop, newest 1000 survive in order. 71 tests pass, clippy clean.

Replying separately on the scope split / `Generating` rename / GIF question.

graykode:
Pushed `bde1597` — caught a second one during smoke testing.

The `< 30s` branch in both collectors was using `!current_task.is_empty()` to split Executing vs Thinking. But `current_task` is empty whenever the latest assistant turn ended on a plain text `end_turn` (no tool_use), so every just-finished idle session rendered as **Thinking** when it should have been Waiting — visible the moment you point abtop at a Claude session that just answered.

Routed Thinking through the actual "model is generating" signal instead:

- Claude: `cached.last_user_ts_ms > 0` — the parser already maintained this (set on user lines, cleared on assistant lines), and `thinking_since_ms` from #67 already uses it for the live Think row in the timeline. Status now reads from the same source of truth, so the two stay in sync.
- Codex: added `model_generating: bool` to `CodexJSONLResult`, set on `user_message` and cleared on `agent_message` / `task_complete`. Same semantics as Claude.

Both collectors now correctly fall back to Waiting in the `< 30s` branch when no tool is running and the model isn't generating. Four tests covering both directions in both collectors. 75 pass, clippy clean.

Still open for your call: scope split, `Generating` rename, GIF.

graykode:
Pushed `f1224e2` — third pass.

The previous fix (route Thinking through `last_user_ts_ms > 0`) wasn't enough: live testing showed sessions still rendering Thinking when the agent was clearly idle. The transcript signal flips on too many "looks active" states that aren't actually computing — tool finished waiting on a permission prompt, session crashed mid-loop with the trailing line being a `user` (tool_result), 97% context sessions parked at the edge.

So I dropped the `< 30s` transcript freshness branch entirely and made status purely CPU-driven:

- `has_active_descendant` → `Executing`
- claude/codex `cpu_pct > 1.0` → `Thinking`
- otherwise → `Waiting`

This is exactly what `stale_status` (the helper you extracted) already did for stale transcripts — now it runs uniformly. Real generating burns CPU on the agent process, so CPU is both necessary and sufficient. The mtime gate was just letting stale transcripts dictate status.

Removed the now-unused `model_generating` field from `CodexJSONLResult` (and the parser updates from the previous commit) plus an `unused_variables` warning in Claude's `load_session`. 73 tests pass, clippy clean. The two new Claude `load_session` tests pin the regression both ways: idle claude → Waiting no matter what the transcript tail says, and `cpu_pct > 1` → Thinking.

graykode:
Pushed `2a9e06c` — fourth and (hopefully) final pass on the Thinking detection.

Pure CPU was wrong too: `ps -o %cpu` is a cumulative lifetime average, so any long-running session that did real work earlier keeps reading >1% even when fully idle. Live testing showed Thinking still firing on agents sitting at the prompt.

The signal that actually matches "model is computing right now" needs **both**:

1. Transcript mtime advanced within the last 5s (claude only writes when it emits something; bursts during a streamed reply touch the file frequently).
2. Trailing transcript line is a user message — `cached.last_user_ts_ms > 0` for Claude, `result.model_generating` for Codex. Without this, a just-finished assistant turn renders Thinking for ~5s every time.

Either alone misfires. Both together pin Thinking to the genuine "user just submitted, model has not replied yet" window. Final decision table for both collectors:

```
has_active_descendant                     -> Executing
mtime < 5s AND trailing_event_is_user    -> Thinking
otherwise                                 -> Waiting
```

Cleanup:
- Removed `collector::stale_status` (no longer called) and its mod.rs tests.
- Re-introduced `model_generating` on `CodexJSONLResult` to mirror Claude's `last_user_ts_ms` semantics.
- Tighter regression tests: `stale_transcript+busy_cpu must still be Waiting`, `recent transcript with trailing user_message must be Thinking`, plus the two parser-level tests for Codex's `model_generating` lifecycle.

72 tests pass, clippy clean. Verified locally on a real session including the case that surfaced the bug (long context session sitting idle while user types — now correctly Waiting).

graykode:
Pushed `2c9b32a` — one more during smoke testing. Status was flickering Think ↔ Wait throughout tool loops because Claude serializes tool results as `user`-role transcript lines, and the last_user_ts_ms update treated those as the start of a new thinking window:

```
assistant(tool_use) -> last_user = 0  (Wait)
user(tool_result)   -> last_user > 0  (Think)   <- spurious
assistant(next)     -> last_user = 0  (Wait)
user(tool_result)   -> last_user > 0  (Think)   <- spurious
```

From the user's perspective it's one continuous turn; the tool_result wrappers are internal plumbing. Detect them by content shape (every block has `type: tool_result`) and skip the `last_user_ts_ms` update for those lines. Mixed-content messages stay treated as real prompts so we never silently drop genuine user input.

Codex doesn't need the same fix — its tool outputs flow through `response_item`, not `user_message`, so `model_generating` is already only set on actual prompts.

73 tests pass, clippy clean. Verified live: status now stays `Think` (or transiently `Exec` when a descendant fires) across the whole tool loop instead of strobing.

graykode (approved):

PR #56, opened 2026-04-20
fix: exclude cache_creation from context window percentage

### LINKED ISSUE #54: Discrepancy between abtop and CC internal counters
Here's a screenshot of abtop on an idle session; i don't understand the 346% context, is that the sum of all percentages across subagents?

Image

And here's my statusline on that session for comparison

Image

tbouquet:
## Summary

Fixes #54 - context percentage showing 346% on idle sessions.

## Root cause

`last_context_tokens` was computed as `input_tokens + cache_read + cache_creation`. The `cache_creation_input_tokens` field represents tokens written to the cache for **future** requests - it's not content in the current context window.

With heavy prompt caching, `cache_creation` can be very large (e.g. 150K tokens), causing the percentage to exceed 100%:

```
Before: (50K input + 150K cache_read + 150K cache_creation) / 200K window = 175%
After:  (50K input + 150K cache_read) / 200K window = 100%
```

## Fix

```diff
- result.last_context_tokens = inp + cr + cc;
+ result.last_context_tokens = inp + cr;
```

Context window = `input_tokens` (fresh) + `cache_read_input_tokens` (reused from cache). This matches what Claude Code's own statusline reports.

## Test plan

- [x] `cargo test` - 35/35 pass (updated assertion in test_parse_transcript_basic_tokens)
- [x] Verified: Claude Code statusline shows 53% on the same session that abtop showed 346%

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -230,7 +230,7 @@ Not provided in data files. Derive:
   - `claude-opus-4-6[1m]` → 1,000,000
   - `claude-sonnet-4-6` → 200,000
   - `claude-haiku-4-5` → 200,000
-- **Current usage**: last `assistant` line's `input_tokens + cache_read_input_tokens + cache_creation_input_tokens`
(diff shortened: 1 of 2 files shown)


graykode (approved):
Thanks for contribution! @tbouquet

PR #58, opened 2026-04-20
feat: discover Claude sessions via PID

cmtonkinson:
This brings Claude Code session discovery into alignment with Codex, the benefit of which is that when Claude Code is being used across multiple profiles, or with a nonstandard profile path(s), it will still be picked up by abtop.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -75,15 +75,26 @@ Panel descriptions:
 
 All read-only from local filesystem + `ps` + `lsof`. No API calls, no auth.
 
-### 1. Claude Code session discovery: `~/.claude/sessions/{PID}.json`
+### 1. Claude Code session discovery: process + config-root mapping
+
+Discovery strategy:
+1. Find running `claude` processes via `ps`
+2. Map PID → open files/directories via `lsof`
+3. Infer Claude config roots from open paths that contain `sessions/` and `projects/`
+4. Read `{config-root}/sessions/{PID}.json`, falling back to scanning session files for the matching embedded PID
+5. Parse `{config-root}/projects/{encoded-path}/{sessionId}.jsonl`
+
+Fallback config roots are still scanned: `~/.claude`, abtop's own `CLAUDE_CONFIG_DIR`, and on Linux any `CLAUDE_CONFIG_DIR` read from `/proc/{pid}/environ`.
+
+Session file format:
 ```json
 { "pid": 7336, "sessionId": "2f029acc-...", "cwd": "/Users/graykode/abtop", "startedAt": 1774715116826, "kind": "interactive", "entrypoint": "cli" }
 ```
 - ~170 bytes. Created on start, deleted on exit.
-- Scan all files, verify PID alive with `ps -p {pid} -o command=` containing `/claude`.
+- Verify PID alive with shared `ps` data containing a `claude` binary.
 - Skip `--print` sessions (abtop's own LLM calls for summary generation).
 
-### 2. Claude Code transcript: `~/.claude/projects/{encoded-path}/{sessionId}.jsonl`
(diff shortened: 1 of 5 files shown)


tbouquet:
Nice work on the PID-first approach - multi-profile discovery is a real gap. A few suggestions:

### Consider `/proc/pid/fd` instead of `lsof`

The current tick loop already spawns `ps` + `lsof` (ports) + `lsof` (Codex). Adding another `lsof` for Claude means 4 fork+exec per tick. On Linux we can get the same data with zero process spawns:

```rust
fn find_open_files(pid: u32) -> Vec<PathBuf> {
    let fd_dir = format!("/proc/{}/fd", pid);
    let Ok(entries) = std::fs::read_dir(&fd_dir) else { return vec![] };
    entries.flatten()
        .filter_map(|e| std::fs::read_link(e.path()).ok())
        .collect()
}
```

This is how btop/htop discover per-process info - ~100x faster than lsof. I just submitted #60 that applies this same approach to the Codex collector, replacing its `lsof` call with `/proc/pid/fd` scanning. Happy to help apply the same pattern here.

### `find_session_file_for_pid` fast path

Claude typically names session files `{PID}.json`. A direct path check before the full directory scan would avoid unnecessary I/O:

```rust
let direct = sessions_dir.join(format!("{}.json", pid));
if direct.exists() { return Some(direct); }
// ...fall back to full scan...
```

### Tests

665 lines of new logic with no tests - `config_dirs_from_lsof_info()` and `find_session_file_for_pid()` would be good candidates for unit tests since they parse external input.

### Split suggestion

Would you consider splitting into two PRs? The structural refactor (`ConfigDir::new`, `merge_config_dirs`, `seen_ids` dedup) is clean and could land independently of the discovery change.

cmtonkinson:
Hey! Thanks for taking a look! I've made some changes:

Optimizations & tests
For performance discovering Claude processes:
* Use `/proc` on Linux
* Use `proc_pidinfo` on macOS
* Fallback to `lsof` only when needed to avoid spawning
* Prefer the direct `{PID}.json` fast path

Deduplicated aggregation, and added a bunch more test coverage.

graykode (approved):
lgtm — security clean: read-only `/proc` + `proc_pidinfo` (libproc syscalls) + `lsof` fallback, no new network/shell. Vetted `proc_pidinfo` 0.1.4 (mmastrac, ~570 LoC, libc FFI only, source matches crates.io tarball). Symlink rejection in `find_session_file_for_pid` and `find_transcript_in_config` is solid and explicitly tested.

Pushed two commits to your branch:
- `a34f5a3` — merge `main` to clear the conflict (Cargo.toml: kept both `apple` and `linux` target deps).
- `f3c72fb` — re-applied a fix that was lost when you force-pushed `Optimizations & tests`. `find_session_file_for_pid` had `?` on per-file read/parse errors inside the loop, so one malformed JSON in the sessions dir would bubble `None` out and abort the whole PID-fallback scan. Replaced with `continue` and added a regression test (`test_find_session_file_for_pid_skips_bad_files_and_continues`).

50/50 tests pass post-fix on macOS (libproc path); Linux `/proc` path will be exercised by CI. Linux runtime testing of the new discovery path hasn't actually happened locally, just the compile — worth noting in case anything proc-fd-related surprises us.

Two small things, neither blocking:
- `is_claude_config_root` accepts any directory with `sessions/` and `projects/` subdirs. Unlikely false positive in real life, but a stray dir matching that shape (e.g. some build cache) would get scanned. The `find_session_file_for_pid` PID-match check downstream means this is harmless — just wasted readdir syscalls.
- `candidate_config_roots_from_path` walks every parent of every open path on every tick. Cheap per call but multiplied by N claude PIDs × M open fds. For a Pro user with one `claude` process this is nothing; if someone has 10+ claudes open it'd be worth memoizing.

PR #59, opened 2026-04-20
fix: summary column truncated despite available space

mquinnv:
## Summary
- Summary column was truncated at 28 chars by `sanitize_fallback()` regardless of available column width
- Changed column constraint from `Min(6)` to `Fill(1)` so the summary actually expands to fill remaining space
- Removed hardcoded "…" ellipsis — ratatui clips to column width naturally
- Added cache purge for old truncated summaries (ending with "…") so they regenerate on next launch
- Bumped LLM summary rejection threshold from 40 to 80 chars

## Test plan
- [ ] Run `cargo build` and verify clean compile
- [ ] Launch `abtop` with active Claude sessions — summaries should fill the available column width
- [ ] Resize terminal — summaries should adapt to column width without pre-truncation
- [ ] Delete `~/Library/Caches/abtop/summaries.json`, restart — old truncated cache entries should be purged and regenerated

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -14,15 +14,10 @@ const MAX_SUMMARY_RETRIES: u32 = 2;
 
 /// Produce a terminal-safe fallback summary from a raw prompt.
 fn sanitize_fallback(prompt: &str, max_len: usize) -> String {
-    let cleaned: String = prompt.chars()
+    prompt.chars()
         .filter(|c| !c.is_control() || *c == ' ')
         .take(max_len)
-        .collect();
-    if prompt.chars().count() > max_len {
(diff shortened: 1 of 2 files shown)


graykode (approved):

graykode:
LGTM. Build/clippy/tests all green on post-merge tree.

- `src/app.rs:16-21` — dropping the `…` suffix loses the truncation indicator when the raw prompt exceeds 80 chars; fine if intentional.
- `src/ui/sessions.rs:267` — `Min(6)` → `Fill(1)` is effectively equivalent here (single flex slot). The real fix is the `sanitize_fallback` change.

PR #60, opened 2026-04-20
perf: replace lsof with /proc/pid/fd for Codex JSONL discovery

tbouquet:
## Summary

On Linux, scan `/proc/{pid}/fd` symlinks directly to find open `rollout-*.jsonl` files instead of spawning `lsof`. Eliminates one fork+exec per tick (~400 syscalls, ~10-30ms) when Codex sessions are active.

Falls back to `lsof` on macOS where `/proc` is not available.

## Before/After

| | Before | After (Linux) |
|--|--------|---------------|
| Method | `Command::new("lsof").args(["-F", "pn", "-p{pid}"])` | `fs::read_dir("/proc/{pid}/fd")` + `read_link` |
| Cost | fork+exec+parse (~10-30ms, ~400 syscalls) | ~20 readlink() calls (~0.1ms) |
| Platform | All | Linux native, macOS fallback to lsof |

## Test plan

- [x] Path matching uses `file_name().to_str()` instead of `to_string_lossy()` for correctness on non-UTF8 paths

### DIFF
--- a/src/collector/codex.rs
+++ b/src/collector/codex.rs
@@ -5,6 +5,7 @@ use std::collections::HashMap;
 use std::fs;
 use std::io::{BufRead, BufReader, Read};
 use std::path::{Path, PathBuf};
+#[cfg(not(target_os = "linux"))]
 use std::process::Command;
 
 /// Collector for OpenAI Codex CLI sessions.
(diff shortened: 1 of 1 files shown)


graykode (approved):
lgtm — security clean (read-only /proc scan, no new deps/network/shell). Build/clippy/tests green; CI's ubuntu-latest job will exercise the Linux path.

A couple of small things worth noting (none blocking):
- Behavior diff vs the lsof path: when a PID has multiple matching `rollout-*.jsonl` open, the new path takes the first FD (`break`), the old path took the last. In practice codex opens one, so it shouldn't matter — just calling it out.
- `fs::read_link` returns whatever the symlink points at; if a codex process were ever induced to open a file like `/etc/foo/rollout-x.jsonl`, abtop would happily parse it. The PIDs are already filtered to codex processes so this is essentially "if codex is compromised, abtop reads what codex told it to," which is the same trust model as the lsof path. Fine to leave as-is.

PR #61, opened 2026-04-20
perf: replace ps spawn with /proc direct reads on Linux

tbouquet:
> **Depends on #60** - apply on top of that PR's branch
> Part 2/3 of Phase 1 /proc optimization (#60 -> this -> next)

## Summary

On Linux, read `/proc/{pid}/stat` + `/proc/{pid}/cmdline` directly instead of spawning `ps -ww -eo pid,ppid,rss,%cpu,command`. Eliminates one fork+exec per tick (~500 syscalls, ~15ms).

Falls back to `ps` on macOS where `/proc` is not available.

## What's parsed from /proc

| File | Fields extracted |
|------|----------------|
| `/proc/{pid}/stat` | ppid, utime, stime, starttime, rss (pages) |
| `/proc/{pid}/cmdline` | Full command (NUL-separated -> space-joined) |
| `/proc/uptime` | System uptime for CPU% calculation |

CPU% is computed as lifetime average `(utime+stime) / elapsed_ticks * 100`. This differs from `ps`'s instantaneous sample but works correctly for abtop's Working/Waiting threshold (`cpu_pct > 1.0`) since long-idle processes see their average decline below 1% over time.

## New shared utility

Introduces `scan_proc_fds(pid) -> Vec<PathBuf>` in `process.rs` - resolves all `/proc/{pid}/fd` symlinks. Used by both this PR's port discovery and #60's Codex JSONL discovery, eliminating duplicated readdir+readlink loops.

## Dependencies

- Adds `libc = "0.2"` (Linux-only, for `sysconf` to get CLK_TCK and page size)

## Test plan

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -14,15 +14,10 @@ const MAX_SUMMARY_RETRIES: u32 = 2;
 
 /// Produce a terminal-safe fallback summary from a raw prompt.
 fn sanitize_fallback(prompt: &str, max_len: usize) -> String {
-    let cleaned: String = prompt.chars()
+    prompt.chars()
         .filter(|c| !c.is_control() || *c == ' ')
         .take(max_len)
-        .collect();
-    if prompt.chars().count() > max_len {
-        format!("{}…", cleaned)
-    } else {
-        cleaned
-    }
+        .collect()
 }
 
 /// Outcome of an Enter-key jump attempt. Distinct from `Option<String>` so
@@ -235,7 +230,7 @@ impl App {
                     *count += 1;
                     if *count >= MAX_SUMMARY_RETRIES {
                         // Exhausted — store sanitized fallback using prompt from worker
-                        self.summaries.insert(sid, sanitize_fallback(&prompt, 28));
+                        self.summaries.insert(sid, sanitize_fallback(&prompt, 80));
                         save_summary_cache(&self.summaries);
                     }
                 }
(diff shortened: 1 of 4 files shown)


graykode (approved):
lgtm — security clean (read-only /proc, libc 0.2 is the canonical FFI crate, the two `unsafe sysconf` calls are inherently safe). Build/clippy/tests green on macOS fallback path; Linux compile is covered by CI's ubuntu-latest.

Noticed while reviewing:
- **Cargo.lock not updated.** Adding `libc` to `Cargo.toml` regenerates the lockfile on first build. Worth committing the lockfile change here so CI doesn't see lockfile churn.
- **CPU% semantic change** is fine for the Working/Waiting threshold (which is mtime-based, not cpu-based) but is a UI-visible diff vs the old `ps` instantaneous sample. You already flagged this in the PR body — just confirming it's an acceptable trade.
- **No unit tests for the `/proc/{pid}/stat` parser**, even though it's the trickiest part (the `rfind(')')` trick around comm with spaces/parens). A couple of `parse_stat()` tests with crafted strings would catch future regressions cheaply.

None of these block merge.

graykode (approved):

graykode:
Closing as superseded by #62.

Since #62 was stacked on this branch, merging #62 already brought all of this PR's content into `main` (libc dependency, `get_process_info` Linux variant, etc.). Confirmed by merging current `main` back into this branch — auto-resolved with zero net diff against `main`.

Thanks for the perf work — this and #62 together gave us the full /proc-based path on Linux.

PR #62, opened 2026-04-20
perf: replace lsof ports with /proc/net/tcp parsing on Linux

tbouquet:
> **Depends on #60 + #61** - apply on top of those PRs
> Part 3/3 of Phase 1 /proc optimization (#60 -> #61 -> this)

## Summary

On Linux, parse `/proc/net/tcp` + `/proc/net/tcp6` for LISTEN sockets and match inodes via `scan_proc_fds()` instead of spawning `lsof -i -P -n -sTCP:LISTEN`. Eliminates the last fork+exec from the port discovery path (~400 syscalls, ~30ms).

Falls back to `lsof` on macOS where `/proc` is not available.

## How it works

1. Parse `/proc/net/tcp[6]` - state `0A` = LISTEN, extract port (hex) and inode
2. Scan `/proc/[pid]/fd` symlinks via `scan_proc_fds()` (shared helper from #61)
3. Match `socket:[inode]` targets to listening port inodes
4. Build pid -> ports map

## Phase 1 complete

With all three PRs merged, the result on Linux:

| Spawn | Before | After |
|-------|--------|-------|
| `ps` | fork+exec every 2s tick | `/proc` direct read |
| `lsof` (Codex) | fork+exec every tick | `/proc/pid/fd` readlink |
| `lsof` (ports) | fork+exec every 5 ticks | `/proc/net/tcp` parse |
| **Total** | ~1500 syscalls, ~50ms | ~200 syscalls, ~5ms |

Remaining spawns: `git status` (slow tick) and `sqlite3` (OpenCode).

## Test plan

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,9 @@ dirs = "6"
 chrono = { version = "0.4", features = ["serde"] }
 tempfile = "3"
 
+[target.'cfg(target_os = "linux")'.dependencies]
+libc = "0.2"
+
 # The profile that 'dist' will build with
 [profile.dist]
 inherits = "release"
--- a/src/collector/codex.rs
+++ b/src/collector/codex.rs
@@ -306,22 +306,13 @@ impl CodexCollector {
         #[cfg(target_os = "linux")]
         {
             for &pid in pids {
-                let fd_dir = format!("/proc/{}/fd", pid);
-                let entries = match fs::read_dir(&fd_dir) {
-                    Ok(e) => e,
-                    Err(_) => continue,
-                };
-                for entry in entries.flatten() {
-                    if let Ok(target) = fs::read_link(entry.path()) {
-                        // Match on the file name component to avoid lossy UTF-8
-                        // conversion issues on the full path.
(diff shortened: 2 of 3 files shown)


graykode (approved):
lgtm — security clean (read-only `/proc/net/tcp[6]` + `/proc/{pid}/fd`, no new deps over #61). Build/clippy/tests green on macOS fallback path; Linux compile covered by CI.

Two small things:
- **Doc comment contradicts the implementation.** `src/collector/process.rs:171` says `Only scans FDs for PIDs in known_pids (from get_process_info)` but the function takes no `known_pids` argument and the inline comment at line 199–202 explicitly says you scan all `/proc` PIDs (which is the right call for orphan-port detection). Just delete the stale claim from the doc comment.
- Same Cargo.lock note as #61 — once that one's lockfile is committed this inherits the fix.

The `0A`/inode parsing matches `proc(5)`, and the `socket:[N]` strip pattern is the right shape. Performance at the 10s slow-tick cadence should be fine even on busier boxes.

graykode (approved):

PR #65, opened 2026-04-21
feat: animate tool timeline bar for currently-running tools

graykode:
cc @tbouquet — pinging since you've been in this area.

## Summary

The transcript parser only learns a tool's duration when the *next* `user` turn arrives, so a currently-running tool (WebSearch, long Bash, Agent spawn) sits at `duration_ms = 0` and renders as an empty bar. Hard to tell from the UI whether abtop is broken or Claude is genuinely still working.

This PR animates the bar in real time for any tool whose assistant turn hasn't closed yet.

## How it works

- `AgentSession` gains `pending_since_ms: u64` — the unix-epoch ms of the open assistant turn, or 0 when nothing is in flight. The collector already tracks this in `TranscriptResult::last_assistant_ts_ms`; it's just plumbed through.
- `draw_timeline` computes a `live_duration` per tool at render time: `now_ms - pending_since_ms` for live tools, `duration_ms` for completed ones.
- Live rows are visually distinct:
  - `●` prefix instead of a space
  - Name pulses BOLD ↔ DIM every 500ms
  - Bar color dimmed
  - Label reads `12.3s…` instead of `12.3s`
  - Header shows `TIMELINE (N calls, Xs, 1 running)`
- The `*` longest-bar marker is suppressed for live rows so a running tool that's briefly the max doesn't steal the badge.

## Caveats

- Bar growth is stepwise, not smooth — abtop's main tick is 2s, so the data refreshes at 2s intervals. The 500ms pulse is independent of that.
- If an assistant turn issues multiple parallel tools, the transcript only records completion at the user turn, so all live tools in that turn share the same elapsed. Accurate for the common single-tool case; an approximation for parallel calls.
- `pending_since_ms > 0` can be set even when the last assistant turn had no `tool_use` (pure text reply). That's harmless — there's no `duration_ms == 0` tool to match against, so nothing renders as live.

## Demo

`--demo` mode got an extra `WebSearch` entry on the first session with `duration_ms = 0` and `pending_since_ms` ~6s in the past, so you can run `cargo run -- --demo` to see the animation without a live Claude session.

## Test plan
- [ ] `cargo run -- --demo` — the `WebSearch` row in session 1 should show a growing bar with a pulsing `●` prefix and `…` suffix
- [ ] `cargo run` with a real Claude session actively running a tool — same animation on the currently-running tool
- [ ] When the tool finishes (user turn arrives), the row should transition to the normal completed style without any layout jump


graykode (commented):
lgtm — security is clean (internal UI/collector plumbing, no new net/shell/deps). pulled it into a worktree, cargo build + clippy + test all pass on the post-main-merge tree.

two tiny things, neither blocking:

- `src/ui/sessions.rs:779` comment says "2-tick (~4s) cycle" but `(now_ms / 500).is_multiple_of(2)` flips every 500ms (1s full cycle). matches the PR description, just the comment is off.
- the feature only animates while an assistant turn with `tool_use` is open and waiting on a user response. the common "Claude is thinking / streaming text" window (no `tool_use` in the JSONL yet) still shows a static timeline — expected given the data source, but probably worth a follow-up for making long thinking phases visible too.

PR #67, opened 2026-04-22
feat: animate a live Think row while the model generates its next turn

graykode:
Follow-up to #65.

## Problem

#65 animates a tool bar while an assistant turn with `tool_use` is awaiting the matching `user` response. The other common live state — model is thinking / streaming text, last transcript line is a `user` turn, no assistant written yet — still rendered a static timeline. In practice this is the state you're in for most of a long turn: Claude is generating tokens but hasn't committed the message to JSONL, so there is nothing for the timeline to animate.

## Approach

Track `last_user_ts_ms` alongside `last_assistant_ts_ms` in the transcript parser. Plumb it to `AgentSession.thinking_since_ms`. When non-zero and the session is alive, append a virtual `●Think` row at the end of the timeline whose bar scales against the longest tool call (capped at full width for long thinks). Pulses bright/dim on a 500ms clock, same as the pending-tool row from #65.

The Think row disappears the moment the next assistant line lands in JSONL — no layout jump, no special teardown.

## Incidental fix

The transcript merge unconditionally copied `last_assistant_ts_ms` from the latest delta. On a no-new-data tick, `parse_transcript` returns early with a zero-initialized delta, which clobbered the cached timestamp and broke the #65 animation between polls. Added a `saw_turn` flag so the timestamps only propagate when the delta actually observed a `user`/`assistant` line.

## Caveats

- Codex collector sets `thinking_since_ms = 0` — parity with `pending_since_ms`. Extending to Codex rollouts is a separate PR.
- \"Thinking\" here means \"last transcript line is a user-type line and no assistant follow-up yet.\" Covers both fresh prompts and tool_result lines, which is what we want — both represent the model working on its next step.
- If the session has been idle for hours with the last line being a user prompt the user never sent more of, we'd still render a growing Think row. Acceptable: the session status (`◌ Wait`) and the duration label itself tell the true story.

## Demo

Session 3 in `--demo` now sets `thinking_since_ms = now - 1s`. `cargo run -- --demo`, navigate to session 3, press `L` to focus — Think row grows for ~2s and caps at 100%.

demo

## Test plan
- [ ] \`cargo run -- --demo\` — session 3 shows a growing \`●Think\` row that caps at 100% width
- [ ] \`cargo run\` against a real Claude session that has just submitted a prompt — Think row appears, disappears when the next assistant message is written
- [ ] \`cargo run\` against a session with a running tool (the #65 case) — pending-tool animation still works, doesn't flicker between 2s ticks (regression test for the merge fix)


graykode (commented):
Security clean — internal UI plumbing only, no network/shell/deps/CI changes, scope matches the description. Build, clippy -D warnings, and all 50 tests pass on the PR tree. The saw_turn merge fix is correct: it flips only inside the user/assistant match arms (summary and other types stay in the _ => {} branch, so a no-op delta keeps saw_turn=false and the cached live markers survive). I also traced the pending_since_ms vs thinking_since_ms interaction — the parser clears last_assistant_ts_ms inside the user handler before setting last_user_ts_ms, so they stay mutually exclusive end-to-end and the UI never has to reconcile both. One drive-by suggestion: the saw_turn guard is exactly the kind of incidental fix that can silently regress, so a tiny unit test (parse_transcript with from_offset pointing past EOF returns saw_turn=false and zero timestamps) would be cheap insurance. Not blocking.

PR #69, opened 2026-04-22
fix: track live session_id after /clear (#68)

### LINKED ISSUE #68: Behavior with /clear between sessions
Hi,
It seems that when you chain sessions with a simple /clear to reset the context, abtop ends up loosing track of what's happening in the session.
Do you confirm that this is somehow expected/normal with the current implementation?
S.

graykode:
## Summary
- Fixes #68. \`/clear\` mints a new \`sessionId\` + new \`{sid}.jsonl\` without rewriting \`sessions/{PID}.json\`, so abtop kept tailing the abandoned transcript and every counter (tokens, context %, current task) froze until the process exited.
- After parsing \`SessionFile\`, look up the most recently modified \`*.jsonl\` in the PID's project dir (filtered to \`mtime >= startedAt\` with a 5s grace) and override \`sf.session_id\` with its stem when it differs. The existing \`transcript_cache\` eviction naturally drops the stale sid once it's no longer in the active set.
- Skipped lsof-based detection: verified on a live PID that Claude Code opens/closes the transcript per append, so the FD isn't reliably visible between polls.

## Test plan
- [x] \`cargo test\` — 58 pass, 4 new (picks newest jsonl, filters by startedAt, empty-dir None, load_session override after /clear)
- [x] \`cargo clippy --all-targets -- -D warnings\` — clean
- [ ] Manual: run abtop, \`/clear\` in a live claude session, confirm counters roll over to the new sid within one tick

### DIFF
--- a/src/collector/claude.rs
+++ b/src/collector/claude.rs
@@ -125,11 +125,14 @@ impl ClaudeCollector {
             }
         }
 
+        let discovery_ctx = build_discovery_context(&session_paths);
+
         let mut sessions = self.load_session_paths(
             &session_paths,
             &shared.process_info,
             &shared.children_map,
             &shared.ports,
+            &discovery_ctx,
         );
 
         // Evict transcript cache for sessions that no longer exist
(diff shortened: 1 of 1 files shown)


graykode:
Addressed the review in ee6d252:

- **Worktree parity:** introduced `resolve_project_dir` that locates the dir holding the original transcript (exact-match first, then worktree sibling scan), and both sid-lookup + transcript path now share that root. Replaces the old `find_transcript_in_config`.
- **Cross-PID hijack:** `DiscoveryContext` pre-computes `pid→claimed_sid` and `cwd→pid_count` per tick. Override is skipped when >1 claude PID shares a cwd; in the single-PID case other PIDs' claimed sids are excluded from live candidates.
- **Tests (+5, 49 total in module):** worktree parity, multi-PID skip, excluded-sids filter, 5s grace window boundary, + counter assertion on the existing /clear test to catch regressions where we'd swap sid but tail the wrong file.

`cargo test` 62/62, `clippy -D warnings` clean.

PR #70, opened 2026-04-22
Fix/clear session transition

smndvgnc:
Added a few more things on top of your changes.
Didn't have time yet to validate on real live test case.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -352,3 +352,4 @@ abtop reads transcripts, prompts, tool inputs, and memory files. These may conta
 - **Terminal size**: minimum 80x24. Panels degrade gracefully when small (context panel hidden first).
 - **PID reuse in port cache**: invalidate cached ports when the set of tracked PIDs changes.
 - **Rate limit staleness**: reject rate limit data older than 10 minutes.
+- **`/clear` + multi-PID same cwd**: after `/clear`, Claude Code mints a new `sessionId` + `.jsonl` without rewriting `sessions/{PID}.json`. abtop overrides the stale sid by picking the newest transcript in the project dir, but this heuristic can't disambiguate ownership when two live `claude` PIDs share a cwd — so the override is disabled in that case and both sessions keep their original sid until exit. Use separate worktrees if live tracking is needed on both simultaneously.
--- a/src/collector/claude.rs
+++ b/src/collector/claude.rs
@@ -265,8 +265,7 @@ impl ClaudeCollector {
 
         // Resolve the project dir that actually holds this session's
         // transcripts. For worktree sessions the on-disk dir does not match
-        // `encode_cwd_path(cwd)`, so locate it via the original sid first
-        // (mirrors the fallback in `find_transcript_in_config`).
+        // `encode_cwd_path(cwd)`, so locate it via the original sid first.
         let project_dir = resolve_project_dir(config, &sf.cwd, &sf.session_id);
 
         // `/clear` mints a new sessionId + transcript without rewriting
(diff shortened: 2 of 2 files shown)


tbouquet:
Ran this locally - build clean, 51/51 tests pass, clippy -D warnings clean. The core shape (live-sid via newest jsonl + DiscoveryContext + cache eviction) is the right call, and the `--print` sibling test is a great catch that would have re-introduced the bug on every machine running abtop. Really nice test coverage.

A few things worth addressing before merge:

### 1. Magic `5_000` in `find_live_session_id`

The grace window is documented in the comment but the literal would be clearer as a named constant. Something like:

```rust
// Clock skew between SessionFile.started_at and the first jsonl flush,
// plus FS mtime granularity (1-2s on some platforms).
const STARTED_AT_GRACE_MS: u64 = 5_000;
```

Makes it greppable and tunable if someone hits a slow-flush edge case.

### 2. `cmd.contains("--print")` duplicated in 3 spots

Lines 225, 306, 890 each do the same substring check. Two of them come from this PR. Worth extracting:

```rust
fn is_print_spawn(cmd: &str) -> bool {
    cmd.contains("--print")
}
```

Two reasons beyond dedup:
- If Claude Code supports `-p` as a short form (I haven't checked), this misses it and the guard silently stops working.
- A single function makes it obvious the predicate must match whatever abtop itself spawns.

### 3. `pids_per_cwd` keyed on `cwd` alone

`DiscoveryContext.pids_per_cwd: HashMap<String, usize>` uses just the cwd string, but `refresh_config_dirs` explicitly supports multiple `CLAUDE_CONFIG_DIR` profiles in the same tick. Two claude processes in the same cwd but under different profiles would false-positive as siblings and suppress the /clear override on both, even though their transcripts live in different trees.

Suggest keying on `(cwd, config_base)` or similar. Not theoretical given the multi-profile plumbing is already there.

### 4. cwd not canonicalized

`sf.cwd` is the raw string Claude Code wrote. Two PIDs with `/home/u/repo` vs `/home/u/repo/` (trailing slash), or a resolved path vs a symlinked one, would count as different cwds in `pids_per_cwd` while referring to the same worktree. A `fs::canonicalize(&sf.cwd).map(|p| p.to_string_lossy().into_owned()).unwrap_or(sf.cwd.clone())` before insertion would tighten this up.

### 5. Semantic shift in `test_resolve_project_dir_rejects_symlinked_matches`

The assertion flipped from `is_none()` to `Some(exact_dir)`. Concretely: when the original transcript is symlinked (which the previous code treated as "do not trust"), we now return the containing dir so `find_live_session_id` can scan inside it. If an attacker planted a symlinked original plus a real malicious `.jsonl` alongside, the live-sid lookup could adopt the attacker's sid.

The attacker already needs write to `~/.claude/projects/`, which is basically game over, so the realistic impact is low - but the commit message doesn't call out that the symlink rejection surface changed. Worth a line in the commit body so future readers don't have to diff the test to notice.

### Nits

- `build_discovery_context` parses every `sessions/{pid}.json`, then `load_session` parses them again a moment later. Caching the parsed `SessionFile` in `DiscoveryContext` would save one parse per PID per tick. Micro.
- `claimed_sids_by_pid: HashMap<u32, String>` is only consumed via `.iter().filter(other_pid).map(sid)`. A `HashSet<String>` of "sids claimed by other PIDs" would match the actual use, unless you're keeping the PID->sid map for a future feature (worth a comment if so).

Nothing blocking. Happy to take a second pass once these are in.

graykode (approved):
lgtm — security clean (no new network/shell/deps, defensive symlink checks throughout). approving.

the `--print` filter in `build_discovery_context` is the load-bearing catch here. without it #69's cross-PID guard would have been silently tripped on every machine running abtop, since the summary spawn lands a `sessions/{PID}.json` in the same cwd as the real session and flips `pids_per_cwd` to 2. so #69 alone would have shipped looking-correct-in-tests but doing nothing in production. nice find.

other things i appreciated:

- `evict_stale_cache()` extraction + the two-poll `test_transcript_cache_evicts_old_sid_after_clear` is exactly the right shape — the `total_input_tokens == 12, not 24` assertion catches the failure mode (counter leakage) directly, not just the symptom.
- the dead-PID branch (`process_info.get(&sf.pid)?`) is a free bonus: stale `sessions/{PID}.json` from crashed sessions stop polluting the sibling count too.
- CLAUDE.md gotcha is honest about the multi-PID-cwd tradeoff. better to document the limitation than to paper over it.

thanks for the thorough follow-up. merging is on me — will squash to keep history tidy.

Issue #45, opened 2026-04-19
Roadmap: agent coverage strategy beyond v0.1


graykode:
## Context

abtop v0.1 was intentionally scoped to Claude Code + Codex CLI, with Gemini/Cursor listed as **Non-Goals** in `CLAUDE.md`. Since then, the project has grown and several contributors have opened PRs adding support for additional agents:

- ✅ **OpenCode** — merged via #40 and released in `v0.4.3`
- 🟡 **Gemini CLI** — #41 (currently Draft, pending this decision)
- 🟡 **Kiro CLI** — #44 (currently Draft, pending this decision)
- 🟡 **Pi** — #63 (candidate for this same policy decision)

Evaluating each new agent ad-hoc isn't fair to contributors (they invest significant work before knowing whether it'll land) and isn't good for the project's long-term shape. Time to define a strategy.

## Current status (2026-05-09)

OpenCode is no longer pending. It is now in core, documented in the README, and available in `v0.4.3`.

OpenCode's supported surface is intentionally narrower than Claude/Codex for now:

- Session discovery from `~/.local/share/opencode/opencode.db`
- Token tracking from local DB totals
- Status detection
- Git status
- Children / ports

Quota/rate-limit gauges remain Claude + Codex only because OpenCode does not expose matching account-level quota data in the same way.

## Questions to answer

1. **What's the inclusion bar?** Criteria could include:
   - User base / popularity
   - Stability of the underlying data source (the agent's session files, DB, or API)
   - Maintenance cost — how often does the agent change its format?
   - Whether abtop's UI abstractions (context %, tokens, rate limits) meaningfully apply

2. **Should core support more agents, or should there be a plugin mechanism?**
   - Pros of in-tree: consistent UX, single binary, easy install
   - Pros of plugins: independent release cadence, lower core maintenance, community can add niche agents without approval
   - A plugin path via the existing `AgentCollector` trait may be feasible

3. **What happens to the `Non-Goals (v0.1)` list in `CLAUDE.md`?**
   - Update it to reflect whatever is decided
   - Decide whether abtop is moving to a broader scope, or staying focused

4. **Release implications** — each new agent is a new data source to keep working across CLI version changes. Need a realistic estimate of the maintenance tail.

## Proposed next step

Keep this issue open for the broader agent coverage policy. OpenCode itself is done, but the general decision still needs to be made before merging additional agent collectors.

- [x] Ship OpenCode in core (#40, `v0.4.3`)
[...]

PR #72, opened 2026-04-24
feat: add hidden_agents config to skip CLIs in TUI

### LINKED ISSUE #71: Possibility to remove Codex from the TUI
I wish I could remove Codex from here

Image

**Use case**
I only use Claude Code, so I don't want to see Codex there

**Additional context**
And some people only use Codex, and would like to hide Claude Code

tbouquet:
## Summary

Closes #71.

Lets users hide specific agent CLIs from the TUI via config file, for people who only use one agent and want a cleaner view.

```toml
# ~/.config/abtop/config.toml
hidden_agents = ["codex"]
```

Match is case-insensitive against the `agent_cli` identifier. Unknown names are ignored so the same config file can be shared across builds with different feature flags (anticipating #45).

## Implementation

- `AppConfig` gains a `hidden_agents: Vec<String>` field.
- The ad-hoc TOML parser in `config.rs` learns to parse single-line string arrays (`["a", "b"]`).
- `MultiCollector::with_hidden(&[String])` replaces `new()` and only pushes collectors whose name is not in the hidden list.
- `App::new_with_hidden` threads the list through; `main.rs` loads the config once and passes it to both `--once` and the interactive paths.

## Follow-up

Once #45 lands and agents are gated by cargo features, a config page toggle (Esc) can be added on top of this same `hidden_agents` key, so users can flip agents on/off at runtime. The TOML contract defined here is designed to be that foundation, no schema change needed for the UI pass.

## Test plan

- [x] `cargo test` passes 81/81, including 8 new tests (`config::tests::*`, `collector::tests::with_hidden_*`)
- [x] `cargo clippy -- -D warnings` clean
- [x] Default behavior unchanged when `hidden_agents` is absent
- [x] Case-insensitive match verified
- [x] Unknown agent names silently ignored

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,17 @@ Colorblind-friendly themes:
 |:-:|:-:|
 | ![deuteranopia](assets/themes/deuteranopia.png) | ![tritanopia](assets/themes/tritanopia.png) |
 
+## Configuration
+
+`~/.config/abtop/config.toml` supports:
+
+```toml
+theme = "btop"
+# Hide specific agent CLIs from the TUI (case-insensitive).
+# Useful if you only use one agent and want a cleaner view.
+hidden_agents = ["codex"]
+```
+
 ## Key Bindings
 
 | Key                | Action                               |
--- a/src/app.rs
+++ b/src/app.rs
@@ -78,7 +78,7 @@ pub struct App {
 }
 
 impl App {
-    pub fn new(theme: Theme) -> Self {
+    pub fn new_with_hidden(theme: Theme, hidden_agents: &[String]) -> Self {
         let (tx, rx) = mpsc::channel();
         let summaries = load_summary_cache();
         Self {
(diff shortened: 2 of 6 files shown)


graykode (approved):
lgtm — security clean (pure config plumbing, no new deps/network/shell). solid test coverage including the nice `rewrite_theme_line` extraction that guards `hidden_agents` from being dropped on theme switches.

one nit, not blocking: the doc on `MultiCollector::with_hidden` says identifiers are matched against each collector's `agent_cli` name, but the implementation actually hardcodes `"claude"` / `"codex"` string literals in `src/collector/mod.rs:123-127`. today they happen to match the `agent_cli` values, but a third collector (or a rename) would need both sites updated in lockstep. a small `fn name(&self) -> &'static str` on the `AgentCollector` trait would make the doc accurate and the filter self-registering. fine as follow-up.

also worth noting: `parse_string_array` splits on bare commas, so `["a,b"]` would mis-parse — non-issue for agent CLI names, just a caveat for anyone reusing the helper later.

PR #63, opened 2026-04-20
Feat: pi agent support

ptahdunbar:
## Summary

   Add `pi-coding-agent` as a first-class session source in abtop.

   This PR adds end-to-end Pi support:
   - discover and parse persisted Pi sessions
   - show Pi rows in the sessions UI
   - include Pi demo data and docs
   - recover idle Pi sessions when the session JSONL fd is closed
   - show temporary pending Pi sessions before the first assistant response is persisted
   - align Pi summary inputs with the existing Claude/Codex summary pipeline

   ## Problem

   abtop supported Claude Code and Codex CLI, but not `pi-coding-agent`.

   In practice that meant:
   - Pi sessions were invisible
   - idle Pi sessions could disappear because Pi does not always keep the JSONL file open
   - new Pi sessions could be missing until the first assistant response was written
   - Pi summary metadata was weaker than Claude/Codex because it did not populate `first_assistant_text`

   ## Solution

   ### Baseline Pi support
   - add `src/collector/pi.rs`
   - register `PiCollector` in `MultiCollector`
   - parse Pi session JSONL under `~/.pi/agent/sessions/...`
   - extract:
     - session id
     - cwd / project
     - model
     - thinking level / effort
     - token totals
     - token history
     - current task
     - child processes / ports

   ### Idle-session recovery
   If Pi is running but no longer has the JSONL file open:
   - inspect process cwd via `lsof`
   - infer the Pi session dir from cwd
   - select the newest matching JSONL in that dir
   - validate the header cwd

   ### Pending-session support
   If the per-cwd Pi session dir exists but no JSONL exists yet:
   - synthesize a temporary `pending-pi-{pid}` session
   - show process-derived metadata until the real JSONL appears

   ### Summary alignment
   Populate `first_assistant_text` for Pi so the shared summary-generation path has the same inputs Claude/Codex already use.

   ### UI / docs
   - add Pi labels/colors in the session table
   - include Pi in demo data
   - document Pi support and pending-session behavior in `README.md` and `CLAUDE.md`

   ## Files

   - `src/collector/mod.rs`
   - `src/collector/pi.rs`
   - `src/collector/process.rs`
   - `src/ui/sessions.rs`
   - `src/demo.rs`
   - `README.md`
   - `CLAUDE.md`

   ## Validation

   ```bash
   cargo test
   cargo clippy -- -D warnings
   cargo run -- --once

PR #44, opened 2026-04-19
feat: add kiro-cli session support

camerondurham:
Adds support for [Kiro CLI](https://kiro.dev) [sessions](https://kiro.dev/docs/cli/chat/session-management/) alongside the existing Claude Code and Codex CLI collectors. Kiro sessions show up in the same sessions table, context window, and tokens panels.

1. New collector discovers live kiro-cli sessions by scanning `~/.kiro/sessions/cli/*.lock`
2. Measure activity by credits (e.g. `115.3cr` for kiro vs keeping existing Claude/Codex `1.2k` / `1.5M` format )  instead of usage rate limits: [kiro plans](https://kiro.dev/pricing/) are monthly credit based and don't have hourly / weekly rate limits currently
1. Kiro reports credits (`cr`) not tokens used, this PR doesn't implement token usage estimation currently but happy to do so

Closes #43

**Question for maintainer**: should we implement token estimation ourselves in this support, or wait until kiro cli emits token statistics in session logs?

Thanks!

## Design decisions

| Decision | Rationale |
|---|---|
| **File-based parsing (JSONL + JSON metadata + `*.lock`), not SQLite** | Kiro's SQLite DB stores session state as a single JSON blob in one TEXT column — no relational rows to query, so SQLite would add a runtime dep for zero parsing wins. Live-session discovery uses `*.lock` files (authoritative, written by kiro-cli). Incremental JSONL tail-parse with offset+inode caching keeps ticks cheap. Mirrors the Claude/Codex pattern. |
| **Match both `kiro-cli` and `kiro-cli-chat` binaries** | kiro-cli is a Bun frontend that spawns `kiro-cli-chat` as the ACP backend. The lock file stores the **backend** PID, so matching only `kiro-cli` misses every live session. |
| **Credits are quota units, not tokens. Treat them separately from Claude/Codex tokens** | Kiro persists real token counts as 0 and uses `metering_usage[].value` (credits) as the activity signal. We route credits ×100 into `total_input_tokens` for storage convenience, but a new `reports_real_tokens()` predicate excludes Kiro from cross-session token aggregations (`Token Rate`, `total`). Per-session rows still show `24.0cr` honestly. The Tokens panel renders a Kiro-specific layout (`Credits / Context% / credits-per-turn sparkline`) since Kiro has no Input/Output/CacheR/CacheW split. If Kiro ever emits real tokens, we prefer them automatically. |
[...]

PR #41, opened 2026-04-18
feat: add Gemini CLI session discovery and monitoring

ericandriol-glitch:
## Summary

Adds `GeminiCollector` implementing the `AgentCollector` trait, enabling abtop to discover and track Google Gemini CLI sessions alongside Claude Code and Codex CLI.

- **Discovery:** finds running `gemini` node processes via `ps`, resolves cwd via `lsof -a -d cwd`, maps to project via `~/.gemini/projects.json`, parses session JSON at `~/.gemini/tmp/{project}/chats/session-*.json`
- **Metrics:** per-message token breakdown (input, output, cached, thoughts), model name, turn count, thinking step subjects as current task, context window estimation
- **UI:** `#GM` label in cornflower blue (#88B0FF), demo session included
- **Tests:** 4 unit tests covering basic parsing, multi-turn aggregation, empty sessions, and project map loading

### Supported features

| Feature           | Gemini CLI |
| ----------------- | :--------: |
| Session Discovery |     ✅      |
| Token Tracking    |     ✅      |
| Context Window %  |     ✅      |
| Status Detection  |     ✅      |
| Current Task      |     ✅      |
| Rate Limit        |     ❌      |
| Git Status        |     ✅      |
| Children / Ports  |     ✅      |
| Subagents         |     ❌      |
| Memory Status     |     ❌      |

Rate limits are not yet exposed in Gemini CLI session files. This can be added when Google surfaces quota data.

### How Gemini CLI stores data

Session files live at `~/.gemini/tmp/{project}/chats/session-{date}-{id}.json`. Each file contains:
- `sessionId`, `startTime`, `lastUpdated`
- `messages[]` with types: `user`, `gemini`, `info`, `error`
- `gemini` messages include `tokens: {input, output, cached, thoughts, tool, total}`, `model`, and `thoughts[]` (thinking steps with subject/description)

### Tested with

- Gemini CLI v0.38.1
- Model: gemini-3-flash-preview
- macOS (Darwin arm64)
- Live session with 242 turns, 48.6M total tokens

## Test plan

- [x] `cargo test` — all 36 tests pass (4 new Gemini tests)
- [x] `cargo build --release` — clean build
- [x] `abtop --once` — correctly discovers live Gemini CLI session alongside 3 Claude Code sessions
- [x] Demo mode includes Gemini session

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #74, opened 2026-04-24
WIP: cargo feature flags per agent collector (demo)

tbouquet:
## WIP / Design discussion - do not merge

This is a **demo branch** showing the cargo-feature-flag mechanism proposed in [#45](https://github.com/graykode/abtop/issues/45) working end-to-end with 6 agent collectors.

Posting as a draft to get your feedback on the approach before opening the real mergeable PRs. Nothing here is meant to land as-is: `test/cargo-features-all-agents` is an integration branch that merges 5 independent feature branches into one so you can see the full picture in one place.

## What's in scope

| Branch | Purpose | Lines |
| --- | --- | --- |
| `feat/cargo-features-agents-init` | Foundation: `[features]` in Cargo.toml, cfg-gate existing collectors (claude + codex) | +107 / -76 |
| `feat/cargo-features-agents-gemini` | Port Gemini collector (#41) behind `gemini` feature | ~720 |
| `feat/cargo-features-agents-pi` | Port Pi collector (#63) behind `pi` feature | ~1100 |
| `feat/cargo-features-agents-kiro` | Port Kiro collector (#44) behind `kiro` feature | ~1200 |
| `feat/cargo-features-agents-opencode` | Port OpenCode collector (#40) behind `opencode` feature | ~480 |

All 5 are feature-flag-gated with `default = ["claude", "codex", "gemini", "kiro", "opencode", "pi"]`. Users who only use one CLI can build a slim binary with `cargo install abtop --no-default-features --features claude`.

## Contributor workflow enabled

A new agent becomes a 3-line diff in wiring:

```rust
// Cargo.toml
default = [..., "myagent"]
myagent = []

// src/collector/mod.rs
#[cfg(feature = "myagent")]
pub mod myagent;
#[cfg(feature = "myagent")]
collectors.push(Box::new(MyAgentCollector::new()));
```

Plus their collector file. Nothing else touched. The per-agent PRs in this stack each touch only ~5 shared lines outside their own collector module.

## What I checked

- `cargo build` default: works
- `cargo build --no-default-features --features <any-subset>`: all combinations I tried build and run
- `cargo clippy -D warnings`: clean on default, claude-only, codex-only, gemini-only, kiro-only, pi-only, opencode-only
- `cargo test`: 112/112 pass (73 init + ports of upstream tests)
- Binary size on Linux x86_64 release: default 2.07 MB, codex-only 1.89 MB (~8% trim), claude-only 2.03 MB (~2% trim)

## Known tradeoffs (not for this PR to resolve)

- Currently six `#[allow(clippy::vec_init_then_push)]` / `#[allow(dead_code)]` annotations, each justified inline. Main cause: some items are owned by one agent but live in shared structs (e.g. `ProcInfo::started_at_ms` read only by Pi; `SharedProcessData::slow_tick` read only by Claude)
- UI label match arms use `#[cfg(feature = "...")]` on each arm. Not pretty but stable since Rust doesn't allow cfg on a whole match
- `reports_real_tokens()` on `AgentSession` still uses a stringly-typed check against `agent_cli == "kiro"`. Would ideally live on the collector trait; kept where it is to avoid a trait refactor in this stack

## What I'm asking you

Before I open the real mergeable PRs (starting with `init`, then the 4 agents as independent PRs), I'd like to know:

1. **Is the feature-flag-per-agent approach acceptable** as the path forward for #45, or would you prefer another shape (all-in-tree, plugin protocol, something else)?
2. **Naming:** `default = [...]` includes all 6 agents. Good default, or should newer agents (gemini/kiro/pi) be opt-in at compile time?
3. **Any blockers** on the generic approach I should fix before splitting this into reviewable PRs?

Happy to adjust or split further. This demo branch can be deleted anytime; the real PRs will be separate branches per agent.

## Not-for-review in this PR

- Any simplify/polish of the ported code (already applied in each sub-branch's second commit)
- The architectural findings from a broader simplify audit of main (flagged separately)


graykode:
Thanks for the proposal. I am going to close this PR because it has drifted too far from `main` and currently has a large conflict surface, which would make review and merge difficult.

If this feature is still needed, please open a fresh PR based on the current `main` branch with the smallest practical scope for the cargo feature flag work. That will be much easier to review and move forward.

PR #80, opened 2026-04-25
feat: host & agent metrics in header, with help (?) and view (v) overlays

tbouquet:
## Summary

Adds two related UX improvements:

### Header metrics
The header now surfaces host vitals (CPU%, MEM%, 1-min load) and an agent aggregate (total RSS across sessions, average context window fill). Layout is width-aware and falls back gracefully on narrow terminals or non-Linux platforms.

```
 abtop vX  CPU 23%  MEM 41%  L 1.8  -  agents Σ620M ctx56%   14:32  3↑ 7●
```

A new `host_info` module reads `/proc/stat`, `/proc/meminfo`, and `/proc/loadavg` directly (no new deps). `AgentAggregate` sums per-session memory and averages context_percent over active agents.

### Shortcut rework
The footer was overflowing with 9+ hint groups. Two new overlays declutter it:

- **`?`** opens a Keybindings overlay grouped by Navigation / Actions / Views / Help. Any key dismisses.
- **`v`** opens a transient View menu collecting toggles previously pinned in the footer (T tree, l timeline, f file audit, 1-5 panels, t cycle theme), each with its current on/off state.

Footer collapses to: `select / jump / kill / filter / view / config / help / quit`.

All previous hotkeys still work outside the overlays — muscle memory preserved.

## Test plan
- [x] `cargo clippy -- -D warnings` (CI invocation)
- [x] `cargo test` (90/90 pass)
- [x] Visually verified via vhs recording (header metrics rendered, ? and v overlays open and dismiss correctly, T toggles tree view)
- [x] Demo mode (`--demo`) populated with synthetic host metrics


graykode (approved):
lgtm — security is clean (no new deps, no network/shell/FS writes outside `/proc`, all `/proc` reads gated to `target_os = "linux"`, scope matches title). a few nits, none blocking:

- `src/ui/header.rs` `pick_metrics`: `base = title.len() + right.len() + 4` uses byte length, but `right` contains `↑` and `●` (3 bytes each in UTF-8). that's ~4 extra "columns" charged against the budget, so host metrics drop out on terminals slightly wider than strictly necessary. swap to `right.chars().count()`.
- `src/ui/header.rs` right-pad: `pad = width - (used + right.chars().count())` budgets 14 chars for the right side, but the rendered spans (`" {now}  "` + `"{active}↑"` + `" {session_count}●"` + `"  "`) total 15. on tight widths the trailing `●` can get pushed off the edge by one column.
- `src/host_info.rs` `HostSampler::sample` docstring says it returns `None` on the first sample, but `sample_cpu` returns `Some(0.0)` on first call, so the first tick on Linux already shows host metrics with CPU=0%. either update the doc or skip the first emission.
- no unit tests for `AgentAggregate::from_sessions` or the host/agent fmt helpers — they're pure functions, would be cheap to add.

footer slimming + `?`/`v` overlays are a nice cleanup. good split into two commits too.

PR #82, opened 2026-04-27
fix(input): close config overlay on second 'c' press

tbouquet:
Noticed c opens the config overlay but a second press of c doesn't close it, you have to hit Esc or q. Both ? and v already toggle on the same key, so adding c to the close arm keeps it consistent.

One-liner in the config_open match.


graykode (approved):
lgtm — security clean (one-line key handler, no new deps/IO). matches the same-key-toggle pattern already used by `Esc | Char('v')` for the view overlay, so consistency win. build + clippy + 90 tests pass.

(minor: the `chore: trigger ci re-run` commit is empty noise — feel free to squash on merge.)

PR #84, opened 2026-04-30
fix(status): treat pending tool_use as Executing

fecet:
Fast tools (e.g. `Bash rm`) finish between CPU samples, so `has_active_descendant` alone misses them and the session reads as Waiting while the agent is still mid-turn. Use `current_task` as the secondary signal: it is cleared on every assistant entry and only repopulated when that turn includes a `tool_use`, so non-empty means a real pending tool — not a text-only reply waiting on the user.

Mirrors the codex collector, which already gates `Executing` on `pending_since_ms > 0`.

Includes regression test `test_load_session_pending_tool_use_is_executing`.

### DIFF
--- a/src/collector/claude.rs
+++ b/src/collector/claude.rs
@@ -478,8 +478,12 @@ impl ClaudeCollector {
         let status = {
             let has_active_descendant =
                 process::has_active_descendant(sf.pid, children_map, process_info, 5.0);
+            // Non-empty current_task = latest assistant turn left a tool_use
+            // unanswered. Catches fast tools (`Bash rm ...`) that finish
+            // between CPU samples, so has_active_descendant alone misses them.
+            let pending_tool = !cached.current_task.is_empty();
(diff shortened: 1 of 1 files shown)


graykode (approved):
lgtm — security clean (pure logic change, no new deps/network/shell/fs).

traced the state model: `current_task` is cleared at the top of every assistant turn (claude.rs:1240) and only repopulated when that turn includes a `tool_use` (claude.rs:1328), so non-empty really does mean "latest assistant turn left a tool unanswered". user/tool_result lines don't touch it, so a fast tool that finishes between samples still reads as Executing until the next assistant turn lands. matches the codex collector's `pending_since_ms > 0` gate. test covers the exact regression. all 91 tests pass.

(unrelated: `cargo clippy --all-targets -- -D warnings` fails on `src/ui/sessions.rs:755` due to `items_after_test_module` — exists on main too, not this PR's problem.)

PR #88, opened 2026-05-01
feat: support multiple Claude Code quota sources

GitZhangChi:
## Summary

Support showing separate Claude Code quota sources when users run multiple Claude profiles via `CLAUDE_CONFIG_DIR`.

abtop already discovers sessions from more than one Claude config root, but quota rendering still assumed one Claude source. With profiles such as `~/.claude-work` and `~/.claude-personal`, each config dir can have its own `abtop-rate-limits.json`; treating all of them as plain `claude` makes the quota panel collapse profile-specific data into a single column.

## Implementation

- Derive a stable Claude quota label from the config directory when the rate-limit file uses the default `source: "claude"`.
  - `~/.claude` stays `claude`.
  - `~/.claude-work` renders as `claude-work`.
- Render quota columns from the available rate-limit sources instead of hard-coding `CLAUDE | CODEX`.
- Treat profiled Claude quota labels as Claude-family sources for rate-limit status promotion.
- Update the statusline setup scripts to:
  - prefer `transcript_path` to infer the active Claude config dir, then fall back to `CLAUDE_CONFIG_DIR`, then `~/.claude`,
  - accept both `five_hour` / `seven_day` and `session` / `weekly` payload shapes,
  - write the rate-limit file atomically.

## Test plan

- [x] `cargo test` (96/96 pass)
- [x] `cargo clippy -- -D warnings`
- [x] `git diff --check`
- [x] Installed locally with `cargo install --path . --force` and verified separate Claude profile rate-limit files are produced/read

### DIFF
--- a/scripts/abtop-statusline.sh
+++ b/scripts/abtop-statusline.sh
@@ -7,23 +7,32 @@
 #
 # Or run: abtop --setup
 
-OUTPUT_FILE="$HOME/.claude/abtop-rate-limits.json"
-
 # Read JSON from stdin
 input=$(cat)
+
+transcript_path=""
+if command -v jq &>/dev/null; then
+    transcript_path=$(echo "$input" | jq -r '.transcript_path // empty' 2>/dev/null)
+fi
+if [[ "$transcript_path" == */projects/* ]]; then
+    CONFIG_DIR="${transcript_path%%/projects/*}"
+else
+    CONFIG_DIR="${CLAUDE_CONFIG_DIR:-$HOME/.claude}"
+fi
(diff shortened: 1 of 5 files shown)


graykode:
Thanks for the thorough write-up and the tests — this is clearly well-considered work. I'm going to pass on this one though. Multi-profile Claude (separate `CLAUDE_CONFIG_DIR` per account on the same machine) is a niche enough use case that I'd rather not pay the ongoing cost: dynamic column count in the quota panel destabilizes the layout, and the same Python parsing logic now lives in two places (`scripts/abtop-statusline.sh` and `src/setup.rs`). Happy to revisit if this pattern shows up more broadly. Appreciate the contribution.

PR #89, opened 2026-05-01
support for windows (not wsl)， tested, works fine on windows11

yangbinbin48:
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,6 +14,7 @@ dependencies = [
  "ratatui",
  "serde",
  "serde_json",
+ "sysinfo",
  "tempfile",
 ]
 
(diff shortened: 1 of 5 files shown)


graykode (approved):
pushed `45c18fa` directly to address the review items below — security pass is clean (sysinfo + a hardcoded `netstat -ano`, nothing new on the network/shell-injection/dotfile front).

what's in that commit:

- long-lived sysinfo `System` (`OnceLock<Mutex<System>>`) for both `get_process_info` and `map_pid_to_sysinfo_open_paths`. `Process::cpu_usage()` is a delta between refreshes, so the previous `System::new()`-per-tick made every Windows process look idle and broke the Working/Waiting threshold.
- `cmd_has_binary` split into `cfg(windows)` / `cfg(not(windows))` impls so the case-insensitive match + `.exe` strip + `\` split don't leak onto linux/macOS, where `Claude` should still not match `claude`.
- new `process::last_path_segment` helper that only splits on `\` under `cfg(windows)`. routed `claude.rs::project_name`, `claude.rs::shorten_path`, and the codex.rs basename calls through it — previously they all silently changed unix path semantics.
- `proc.name()` fallback when `proc.cmd()` is empty on Windows, plus an empty-command skip so we don't insert phantom entries.
- dropped the PowerShell `read_env_var_from_proc` impl. function name promised an env var read, but it was actually `Get-CimInstance Win32_Process` and parsing `--config-dir` out of the command line — spawned powershell every poll tick. folded back into the shared non-linux stub; can revisit per-process env reads as a separate, properly named feature.

`cargo build && cargo clippy --all-targets && cargo test` clean on macOS (91 tests pass). i didn't run on Windows; relying on your earlier verification + the typed API.

approving so this can land. follow-ups separate from this PR:
- #83 closed as duplicate.
- planning to take dist-workspace.toml Windows targets + PowerShell installer + README Windows section from #90 separately (Windows release artifacts), and #90's Arch PKGBUILD as another standalone PR.

thanks for the work — Codex/Claude on Windows is going to land in good shape because of this.

PR #91, opened 2026-05-02
feat(ui): add toggle for projects panel (#87)

graykode:
## Summary

Closes #87. Adds a toggle for the projects panel, assigned to key \`4\`. Ports moves to \`5\`, sessions to \`6\` so the panel-number badges in each block title (¹²³⁴⁵⁶) stay contiguous and match the hotkeys.

The projects panel was the only mid-row panel without a toggle. Users working outside git repos (or hitting the \`✓clean\` glyph rendering issue with fonts like Sarasa Mono K) had no way to hide it.

## Changes

- \`src/app.rs\`: add \`show_projects\` (default \`true\`), shift \`toggle_panel\` and \`config_toggle_selected\` matches, bump \`config_item_count\` 6 → 7.
- \`src/main.rs\`: hotkey range \`'1'..='5'\` → \`'1'..='6'\` (both view-overlay and main key paths).
- \`src/ui/view_menu.rs\`, \`src/ui/config.rs\`, \`src/ui/help.rs\`: surface the new toggle, shift labels.
- \`src/ui/projects.rs\`, \`src/ui/ports.rs\`, \`src/ui/sessions.rs\`: update the panel-number badges (\`projects\` gets \`⁴\`, \`ports\` \`⁵\`, \`sessions\` \`⁶\`).
- \`src/ui/mod.rs\`: gate the projects panel on \`show_projects\`, and let the mid row collapse entirely when all four mid panels are off.

## Test plan

- [x] \`cargo build\`
- [x] \`cargo test\` (90 passed)
- [x] \`cargo clippy --all-targets\` (no new warnings; pre-existing \`items_after_test_module\` in \`src/ui/sessions.rs:755\` is unchanged)
- [x] Verified hotkeys \`1\`–\`6\` toggle the matching panels and that toggling all four mid panels off collapses the row.

PR #92, opened 2026-05-02
feat(config): persist panel visibility across restarts

graykode:
## Summary

Stacked on #91. Panel toggles (1–6, or via the \`c\` overlay) now persist to \`~/.config/abtop/config.toml\` so a layout the user picked survives a restart.

## Changes

- \`src/config.rs\`:
  - New \`PanelVisibility\` struct with one bool per panel; \`AppConfig\` carries it. Missing or malformed keys fall back to \`true\` so an empty config still renders the full UI.
  - Parse \`show_context\` / \`show_quota\` / \`show_tokens\` / \`show_projects\` / \`show_ports\` / \`show_sessions\` from \`config.toml\`.
  - Refactor \`rewrite_theme_line\` into a generic \`rewrite_kv_lines(content, &[(key, value)])\` so both \`save_theme\` and the new \`save_panel_visibility\` share one preserve-unknown-lines path. Comments, \`hidden_agents\`, and any future keys survive both kinds of saves.

- \`src/app.rs\`:
  - \`App::new_with_config\` now takes a \`PanelVisibility\` and seeds the six \`show_*\` flags from it.
  - \`toggle_panel\` and \`config_toggle_selected\` call a new \`persist_panel_visibility\` helper after each change. Save failures surface in the existing transient status footer; success is silent (the panel disappearing/appearing is the visible feedback).
  - \`new_with_hidden\` removed (only one call path remained, and bundling visibility into the constructor signature is cleaner than two near-duplicate helpers).

- \`src/main.rs\`: thread \`cfg.panels\` from \`load_config\` into both the \`--once\` and TUI app constructors.

## Notes

- Stacked on #91 — review that one first; this PR's base will auto-update to \`main\` once #91 lands.
- Saves run on every toggle (atomic \`fs::write\` of a small TOML file). No measurable cost on local disk; the existing \`save_theme\` already uses the same write pattern.

## Test plan

- [x] \`cargo build\`
- [x] \`cargo test\` (92 passed, including 2 new ones for \`rewrite_kv_lines\` multi-key behavior and \`parse_bool\`)
- [x] \`cargo clippy --all-targets\` (no new warnings)
- [x] Manually verified: toggle \`4\` (projects) off, quit, restart → projects stays hidden. Toggle back on → restored after restart.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -90,7 +90,11 @@ pub struct App {
 }
 
 impl App {
-    pub fn new_with_hidden(theme: Theme, hidden_agents: &[String]) -> Self {
+    pub fn new_with_config(
+        theme: Theme,
+        hidden_agents: &[String],
+        panels: crate::config::PanelVisibility,
+    ) -> Self {
         let (tx, rx) = mpsc::channel();
         let summaries = load_summary_cache();
         Self {
@@ -111,12 +115,12 @@ impl App {
             status_msg: None,
             kill_confirm: None,
             theme,
-            show_context: true,
-            show_quota: true,
-            show_tokens: true,
-            show_projects: true,
-            show_ports: true,
-            show_sessions: true,
+            show_context: panels.context,
+            show_quota: panels.quota,
+            show_tokens: panels.tokens,
+            show_projects: panels.projects,
+            show_ports: panels.ports,
+            show_sessions: panels.sessions,
             config_open: false,
             config_selected: 0,
             tree_view: false,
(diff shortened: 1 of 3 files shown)


graykode (commented):
Independent codex pass (second-opinion review): clean.

- security: local fs only, no new deps.
- `rewrite_kv_lines`: preserves non-matching lines, replaces matching keys, appends missing keys. tests `rewrite_theme_preserves_arbitrary_unknown_keys`, `rewrite_theme_appends_when_missing`, `rewrite_panels_replaces_existing_and_appends_missing` cover those paths.
- round-trip: `load_config` parses `show_*`, `App::new_with_config` seeds flags, `toggle_panel` / `config_toggle_selected` call `persist_panel_visibility` → `save_panel_visibility`. holds by inspection.
- `cycle_theme` early-return in `config_toggle_selected`: theme save unaffected (it has its own `save_theme` call).
- non-atomic `fs::write`: same pattern as the pre-existing `save_theme`, no new risk.
- minor non-blocking gap: inline comments on rewritten key lines aren't preserved — `key = value  # note` becomes `key = value`. Worth noting if anyone ever ships a config.toml with inline comments.
- scope: nothing unrelated.

cargo build / test (92 passed) / clippy clean locally.

PR #94, opened 2026-05-02
process: match Claude Code 2.x autoupdater binary layout

### LINKED ISSUE #93: Claude session detection broken on macOS for Claude Code 2.x autoupdater layout
## What's wrong

On macOS, abtop's session list misses all running Claude sessions. The `quota` panel still shows the `CLAUDE` column (because the StatusLine hook writes to a known file path), but the bottom session table only shows `codex` entries — even when multiple Claude Code sessions are clearly running.

## Repro

Tested on macOS 24.6.0 (Darwin 24.6.0), abtop **v0.4.0**, Claude Code **2.1.121**.

```sh
$ ps -p $(pgrep -f '/claude/versions/') -o command=
/Users/a/.local/share/claude/versions/2.1.121 --allow-dangerously-skip-permissions

$ abtop --once
abtop — 3 sessions
  46305 workspace(...)  ◌ Wait gpt-5.5    ...
  22926 workspace(...)  ◌ Wait gpt-5.5    ...
  12813 workspace(...)  ◌ Wait gpt-5.5    ...
```

Five Claude PIDs are alive, but none appear.

## Root cause

Claude Code 2.x ships an autoupdater layout where the running executable lives at `~/.local/share/claude/versions/<version>` — i.e. **the binary file is named after the version string** (`2.1.121`), not `claude`. `~/.local/bin/claude` is just a wrapper.

abtop's `cmd_has_binary` in `src/collector/process.rs` only checks the basename:

```rust
pub fn cmd_has_binary(cmd: &str, name: &str) -> bool {
    let mut tokens = cmd.split_whitespace().take(2);
    tokens.any(|tok| {
        let base = tok.rsplit('/').next().unwrap_or(tok);
        base == name
    })
}
```

For the 2.x layout, `base` is `2.1.121` — not `claude` — so every call from `claude.rs` (`cmd_has_binary(&info.command, "claude")`) returns false, and abtop treats the processes as non-Claude. Sessions are then never matched to live PIDs.

## Suggested fix

Extend the matcher to also accept the autoupdater path shape `<...>/<name>/versions/<filename>`. Conservative — won't over-match (a generic substring or a `claude` dir-segment check would catch unrelated paths).

```rust
pub fn cmd_has_binary(cmd: &str, name: &str) -> bool {
    let mut tokens = cmd.split_whitespace().take(2);
    tokens.any(|tok| {
        let base = tok.rsplit('/').next().unwrap_or(tok);
        if base == name {
            return true;
        }
        let parts: Vec<&str> = tok.rsplit('/').collect();
        parts.len() >= 3 && parts[1] == "versions" && parts[2] == name
    })
}
```

Same shape for the Windows variant. With this change, abtop on macOS shows **8 sessions** (5 Claude + 3 codex) instead of 3.

I have a working patch on a local fork; happy to send a PR.

## Side note (not part of this fix)

While digging, I tried adding a macOS implementation for `read_env_var_from_proc` (currently a stub on non-Linux) using `ps eww` / `KERN_PROCARGS2` — both unreliable: macOS truncates env block output to ~120 chars for non-root callers, so `CLAUDE_CONFIG_DIR` is rarely visible. The existing libproc open-FD inspection in `discover_active_session_paths` already partially compensates, so this isn't a blocker — but worth flagging if anyone wants per-process profile-dir discovery on macOS, it'll need a different mechanism (proc info + entitlements, or the process telling abtop where it lives).

Eridanus117:
Fixes #93.

`cmd_has_binary("claude")` only checked the binary basename. Claude Code 2.x runs the version-named binary directly (e.g. `~/.local/share/claude/versions/2.1.121`), so the basename is the version string — detection fails and live Claude sessions never appear in the session list on macOS.

Extend the matcher to also accept the autoupdater layout `<...>/<name>/versions/<file>`. Conservative pattern: a generic substring match or a dir-named-`claude` check would over-match unrelated paths. Same change applied to the Windows variant for consistency.

A second pass at adding macOS env discovery (`KERN_PROCARGS2` / `ps eww`) was tried and abandoned — the kernel truncates the env block to ~120 chars for non-root callers, so the read is unreliable. The existing libproc open-FD inspection in `discover_active_session_paths` already partially compensates. The non-Linux `read_env_var_from_proc` stub is now annotated with a comment explaining this for the next person who looks.

## Verification

```
$ cargo test --release
test result: 96 passed; 0 failed
```

(3 new tests in `src/collector/process.rs::tests` covering basename match, autoupdater layout, and a non-overmatch guard.)

On a Mac with 5 Claude Code 2.1.121 sessions running:

| | sessions visible |
|---|---|
| before | 3 (codex only) |
| after  | 8 (5 Claude + 3 codex) |

### DIFF
--- a/src/collector/claude.rs
+++ b/src/collector/claude.rs
@@ -1703,6 +1703,13 @@ fn read_env_var_from_proc(pid: u32, var_name: &str) -> Option<String> {
 /// without elevated privileges, so per-process `CLAUDE_CONFIG_DIR` overrides
 /// can't be detected — abtop's own env (resolved in `refresh_config_dirs`)
 /// is the only signal there.
+///
+/// On macOS, `ps eww`/`KERN_PROCARGS2` are unreliable: the kernel truncates
+/// the env block to ~120 chars for non-root callers, so `CLAUDE_CONFIG_DIR`
+/// is rarely visible. Discovery of profile sessions instead piggybacks on
+/// `libproc` open-FD inspection (see `discover_active_session_paths`), which
+/// reads the actual session-file paths a Claude process has open and infers
+/// the config dir from there.
 #[cfg(not(target_os = "linux"))]
 fn read_env_var_from_proc(_pid: u32, _var_name: &str) -> Option<String> {
     None
--- a/src/collector/process.rs
+++ b/src/collector/process.rs
@@ -368,12 +368,12 @@ pub fn last_path_segment(s: &str) -> Option<&str> {
 pub fn cmd_has_binary(cmd: &str, name: &str) -> bool {
     let mut tokens = cmd.split_whitespace().take(2);
     tokens.any(|tok| {
-        let base = tok.rsplit('/').next().unwrap_or(tok);
+        let mut iter = tok.rsplit('/');
+        let base = iter.next().unwrap_or(tok);
         if base == name {
             return true;
         }
-        let parts: Vec<&str> = tok.rsplit('/').collect();
-        parts.len() >= 3 && parts[1] == "versions" && parts[2] == name
+        matches!((iter.next(), iter.next()), (Some("versions"), Some(parent)) if parent == name)
     })
 }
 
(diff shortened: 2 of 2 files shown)


graykode (approved):
lgtm — security clean (pure function + doc-only), 96 tests pass on macOS, the new ones cover the autoupdater path plus two over-match guards. logic is correct: the fallback only fires when basename match fails AND the path has the exact `<name>/versions/<file>` shape, so claude-launch / sibling dirs / unrelated `versions/` paths don't get pulled in.

repro on the PR description (3 → 8 sessions on a 5-claude box) is the right kind of validation.

one tiny nit, non-blocking: `let parts: Vec<&str> = tok.rsplit('/').collect();` allocates per token per tick. since you only need positions 1 and 2, an iterator chain would skip the alloc:

```rust
let mut iter = tok.rsplit('/');
let _file = iter.next();
matches!((iter.next(), iter.next()), (Some("versions"), Some(n)) if n == name)
```

happy to merge as-is though — the alloc is sub-microsecond and the current form reads more clearly. nice fix, thanks for digging into the env-discovery dead-end and leaving the comment so the next person doesn't repeat it.

Issue #95, opened 2026-05-03
codex mcp-server breaks 1-PID-1-session assumption (multi-rollout fd holding)


Eridanus117:
## Problem

`codex mcp-server` is a long-running JSON-RPC service that handles multiple Codex conversations in-process. Each tool call opens a new `rollout-*.jsonl` and **keeps the fd open** (likely a deliberate choice — see (c) below), so a single PID accumulates N rollouts over its lifetime.

abtop's collector assumes 1 PID = 1 rollout (`HashMap<u32, PathBuf>` insert in `map_pid_to_jsonl`, `src/collector/codex.rs`). The consequences:

- Only the last rollout returned by lsof survives the HashMap overwrite. The other N-1 are silently lost.
- Lost rollouts fall through to the "recently finished sessions" pass: if mtime < 5min → shown as PID=0 Done (misleading — the thread is owned by an alive mcp-server); if mtime > 5min → filtered out entirely.
- Idle mcp-server (no rollout opened yet) is never displayed at all, since the active-pass loop only iterates `pid_to_jsonl`.

## Evidence (macOS)

```
$ lsof -p 17870 | grep rollout
codex-aar 17870 ... rollout-2026-05-03T01-18-02-...jsonl
codex-aar 17870 ... rollout-2026-05-03T01-21-40-...jsonl
codex-aar 17870 ... rollout-2026-05-03T01-23-13-...jsonl
codex-aar 17870 ... rollout-2026-05-03T01-26-10-...jsonl
```

mtime spread for those 4 fds: 1h35m → 53m ago. The earliest fd points to a rollout the mcp-server hasn't written to in over an hour but hasn't released either.

A second mcp-server on the same box was holding 5 rollouts spanning 2h22m → 37m. So fd-presence alone overcounts heavily; an "active thread" signal needs an mtime threshold (e.g. < 30s).

Process tree:
```
claude (PID 17533) → codex mcp-server -c profile=qwen36-litellm (PID 17870)
```

This is `claude-code` invoking `codex` as an MCP server for cheap-lane routing — an increasingly common pattern.

## Why this matters

`codex mcp-server` is becoming a normal way to wire Codex into Claude Code (cheap-lane delegation, stateful bridges, multi-agent orchestration via MCP). The 1-PID-1-rollout assumption inherited from Codex CLI / Claude Code interactive doesn't fit.

## Directions I see

I lean toward (b) but want your read before sending a PR.

[...]

PR #96, opened 2026-05-03
Detect codex mcp-server processes in a dedicated MCP panel

Eridanus117:
Closes / refs #95.

## What this does

Adds a process-driven **MCP servers** panel that surfaces every running `codex mcp-server` and the rollouts it currently holds open. By default, mcp-server-owned rollouts are excluded from the sessions panel, fixing the PID=0 "Done" ghosts and the silent-overwrite drop documented in #95.

## Disclosure: which option from #95 I implemented and why

#95 listed three rough directions; this PR implements **(b)** — the new process-driven panel.

I chose (b) over (a) because (a) — making the sessions panel handle 1 PID → N rollouts — would have broken every PID-scoped operation in abtop:
- `x` SIGKILL would kill all in-process threads at once with no UI signal of the connection
- `mem_mb` would double-count across rows that share a PID
- `children`/ports would duplicate to every row that shares a PID
- status detection would be permanently wrong because the PID never dies and finished threads keep their fds open

I deliberately did **not** fix the multi-rollout `HashMap<u32, PathBuf>` overwrite in `CodexCollector::map_pid_to_jsonl`. That fix only makes sense in option (a), and (a) hits the dysfunction above. If you'd prefer (a) or a hybrid, I'm happy to rework — flagging it explicitly here so the choice is on the table rather than buried in a diff.

I did not pursue (c) (push fd-recycling into codex). Keeping fds open is plausibly an intentional codex design choice for thread resume, not a bug — see the discussion in #95.

## Behavior

- New `collector::mcp` module detects `codex mcp-server` processes from the shared `ps` snapshot. Per PID, it lists every open `rollout-*.jsonl` fd (Linux: `/proc/{pid}/fd`; macOS: `lsof`; Windows: `sysinfo`).
- "Active" rollouts use a 30s mtime threshold (`ACTIVE_MTIME_SECS`). Fd presence alone overcounts: on my box, two mcp-servers each held 4–5 rollouts spanning hours, only the most recent of which was actually being written to.
- Panel columns: parent CLI (`claude`/`codex`/`?`), `-c profile=` value (or `default`), `active/total` rollout count, age of latest rollout mtime. Idle mcp-servers (no rollouts) show `0/0 —` so users can still see "Claude has launched the cheap MCP".
- `CodexCollector` skips mcp-server PIDs in its active pass and skips mcp-owned rollouts in the recently-finished pass, removing the PID=0 ghost rows. Pass-through via `SharedProcessData` so the sets are computed once per tick.
- Toggle `7` shows/hides the panel (added to view menu, config overlay, persisted in `~/.config/abtop/config.toml`).
- Toggle `M` flips mcp-server-owned rollouts back into the sessions panel for debugging — with the existing 1-of-N HashMap-overwrite caveat unchanged.
- `--once` prints an MCP server section.

## Verified locally

```
$ abtop --once
abtop — 8 sessions, 14 mcp servers

  mcp pid=17870 parent=claude profile=qwen36-litellm   active=0/4 last=1h
  mcp pid=30022 parent=claude profile=default          active=0/5 last=54m
  mcp pid=46305 parent=claude profile=default          active=0/4 last=9h
  ...
```

`cargo test` passes (103 tests, including 6 new ones for `mcp::detect`, profile parsing, parent-CLI resolution, and active-mtime threshold). `cargo clippy --all-targets` shows only the pre-existing `items_after_test_module` warning in `src/ui/sessions.rs` that's already on `main`.

## Files

- `src/collector/mcp.rs` (new) — detection + multi-rollout fd mapping
- `src/ui/mcp.rs` (new) — panel renderer
- `src/collector/mod.rs` — wires McpDetection into MultiCollector + SharedProcessData
- `src/collector/codex.rs` — sessions-panel suppression hooks
- `src/app.rs` — `show_mcp` + `mcp_servers` + `mcp_suppress_sessions` + toggles
- `src/config.rs` — `show_mcp` persistence
- `src/main.rs` — `1`-`7` panel range, `M` suppression toggle, `--once` MCP section
- `src/ui/{mod,view_menu,help}.rs` — register panel + help/menu entries

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1,4 +1,4 @@
-use crate::collector::{MultiCollector, read_rate_limits};
+use crate::collector::{McpServer, MultiCollector, read_rate_limits};
 use crate::host_info::{AgentAggregate, HostMetrics, HostSampler};
 use crate::model::{AgentSession, OrphanPort, RateLimitInfo, SessionStatus};
 use crate::theme::Theme;
@@ -69,6 +69,14 @@ pub struct App {
     pub show_projects: bool,
     pub show_ports: bool,
     pub show_sessions: bool,
+    pub show_mcp: bool,
+    /// MCP servers detected on the most recent tick (sourced from
+    /// MultiCollector). Populated regardless of `show_mcp` so panel
+    /// toggling doesn't cost a discovery roundtrip.
+    pub mcp_servers: Vec<McpServer>,
+    /// When true (default), mcp-server-owned rollouts are hidden from
+    /// the sessions panel. Toggle with Shift+M.
+    pub mcp_suppress_sessions: bool,
     pub config_open: bool,
     pub config_selected: usize,
     pub tree_view: bool,
@@ -97,6 +105,8 @@ impl App {
     ) -> Self {
         let (tx, rx) = mpsc::channel();
         let summaries = load_summary_cache();
+        let mut collector = MultiCollector::with_hidden(hidden_agents);
+        collector.set_mcp_suppress(true);
         Self {
             sessions: Vec::new(),
             selected: 0,
@@ -105,7 +115,7 @@ impl App {
             rate_limits: Vec::new(),
             prev_tokens: HashMap::new(),
             rate_limit_counter: 5,
-            collector: MultiCollector::with_hidden(hidden_agents),
+            collector,
             summaries,
             pending_summaries: HashSet::new(),
             summary_retries: HashMap::new(),
@@ -121,6 +131,9 @@ impl App {
             show_projects: panels.projects,
             show_ports: panels.ports,
             show_sessions: panels.sessions,
+            show_mcp: panels.mcp,
+            mcp_servers: Vec::new(),
+            mcp_suppress_sessions: true,
             config_open: false,
             config_selected: 0,
             tree_view: false,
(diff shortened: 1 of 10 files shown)


graykode (approved):
nice piece of work. security clean — no new network/deps/shell-injection surface; the only new spawn is `lsof -F pn -p<pid>` with hardcoded argv, paths are stat'd via `metadata` not exec'd, and `is_rollout_path`'s file_name pattern means a path that survives the filter still only gets size+mtime read. 103 tests pass on macOS, the 6 new ones cover detection / profile parsing / parent-cli resolution / non-overmatch / mtime threshold.

i agree with picking option (b) over (a) — the four-way breakage you listed (sigkill, mem double-count, ports duplication, never-dies-status) is real and would have been a worse user experience than the panel split. the explicit disclosure of why you skipped the multi-rollout `HashMap<u32, PathBuf>` fix is exactly the right call for this scope.

a few small things — none blocking:

- `src/app.rs:108` — the explicit `collector.set_mcp_suppress(true)` is redundant since `MultiCollector::with_hidden` already defaults `mcp_suppress: true` (`collector/mod.rs:169`). harmless but can be dropped.
- `src/collector/mcp.rs:140` — `!cmd.contains("grep")` is defense for a case that doesn't apply: `process_info` is built from the system process enumeration, not from `ps | grep` output. fine to keep, but maybe drop the `grep` part of the comment so future readers don't think we're parsing `ps | grep`.
- `src/collector/mcp.rs:164-177` — `parse_profile_flag` substring-matches the literal `"profile="`, which would also match e.g. a `--system-prompt 'profile=foo'` arg if anyone wired one in. extremely unlikely today, just noting since the helper is otherwise fine.
- `src/collector/mod.rs:212` — `mcp::detect` runs every tick, including the path that does `/proc/{pid}/fd` scanning on linux per detected mcp pid. matches the existing `CodexCollector::map_pid_to_jsonl` cadence so it's consistent, but worth keeping in mind if someone's running 50+ mcp servers.
- one ux question for you to decide: when suppression is on and a codex session is mid-turn through an mcp-server, the session panel won't show that turn (it's filtered as mcp-owned). users still see "active=1/N" in the mcp panel, but they lose token tracking + current task for that thread. is that the intended trade-off, or worth a follow-up that promotes the active rollout into a session row keyed off the mcp pid? happy either way, just flagging since the PR description focused on the "ghost Done" case.

approving so you can land. if you want to address the nits in this PR i'll re-stamp; otherwise feel free to merge as-is.

PR #98, opened 2026-05-05
fix(windows): resolve Codex CLI session discovery on Windows

AntaresGG:
## Summary

- Fix `map_pid_to_jsonl` Windows branch to scan `~/.codex/sessions/YYYY/MM/DD/` instead of the process CWD (which never contains rollout files)
- Fix `cmd_has_binary` Windows variant to strip `.js`/`.sh`/`.py` extensions so `node.exe ... codex.js` correctly matches `codex`

## Problem

On Windows, abtop could never discover active Codex CLI sessions due to two bugs:

1. **Wrong scan directory**: The Windows `map_pid_to_jsonl` used `sysinfo` to get the process CWD, then scanned it for `rollout-*.jsonl`. But Codex stores JSONL files in `~/.codex/sessions/YYYY/MM/DD/`, not in the project working directory. Linux works because `/proc/{pid}/fd` resolves open file symlinks regardless of path.

2. **Process name mismatch**: `cmd_has_binary("codex")` failed because Codex CLI runs as `node.exe ... codex.js`. The Windows variant only stripped `.exe`, so basename `codex.js` ≠ `codex`.

## Changes

- `src/collector/codex.rs`: Replace CWD-based scanning with `today_session_dir()` scanning, assigning JSONL files to PIDs by modification time
- `src/collector/process.rs`: Add `.js`, `.sh`, `.py` extension stripping on the Windows `cmd_has_binary` path

## Test plan

- [x] `cargo test collector::codex` — all 14 tests pass
- [x] `cargo test collector::process` — all 3 tests pass
- [x] `cargo build --release` — compiles cleanly on Windows 11
- [x] `abtop --once` shows both Claude Code and Codex CLI sessions simultaneously
- [x] Manual TUI verification (`abtop`) with Codex actively running

### DIFF
--- a/src/collector/codex.rs
+++ b/src/collector/codex.rs
@@ -323,6 +323,10 @@ impl CodexCollector {
     /// equivalent of lsof for enumerating open file descriptors.
     /// Falls back to lsof on macOS/other platforms.
     fn map_pid_to_jsonl(pids: &[u32], sessions_dir: &Path) -> HashMap<u32, PathBuf> {
+        // sessions_dir is consumed only by the windows arm below.
+        #[cfg(not(target_os = "windows"))]
+        let _ = sessions_dir;
+
         let mut map = HashMap::new();
         if pids.is_empty() {
             return map;
--- a/src/collector/codex.rs
+++ b/src/collector/codex.rs
@@ -52,7 +52,7 @@ impl CodexCollector {
             &shared.mcp_server_pids,
         );
         let just_pids: Vec<u32> = codex_pids.iter().map(|(p, _)| *p).collect();
-        let pid_to_jsonl = Self::map_pid_to_jsonl(&just_pids);
+        let pid_to_jsonl = Self::map_pid_to_jsonl(&just_pids, &self.sessions_dir);
         let pid_is_exec: HashMap<u32, bool> = codex_pids.into_iter().collect();
 
         let mut sessions = Vec::new();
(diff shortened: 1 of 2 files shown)


graykode (changes requested):
security clean (pure platform-specific path-scan refactor, no new deps/network/shell). holding off on approval because `cargo clippy -- -D warnings` fails on linux and macOS — `sessions_dir` is only consumed inside the windows `cfg` arm, so on every other platform it's an unused variable. minimal fix:

```rust
fn map_pid_to_jsonl(pids: &[u32], sessions_dir: &Path) -> HashMap<u32, PathBuf> {
    #[cfg(not(target_os = "windows"))]
    let _ = sessions_dir;
    ...
}
```

(or rename to `_sessions_dir`, but the cfg-gated `let _ =` keeps the public-ish signature readable).

the windows logic is fine for the typical single-codex case. two things worth flagging that aren't blockers:
- the mtime-based PID↔file assignment is a heuristic — with two concurrent codex processes whose JSONL files were last touched in different orders, you'll get the wrong file mapped to the wrong PID. the inline comment already acknowledges this so I won't ask you to redesign it, but it's worth keeping in mind.
- `today_session_dir` only walks today's `YYYY/MM/DD` subdir, so a session that started yesterday and is still running across midnight won't be discovered. minor.

`cmd_has_binary` extension stripping looks correct for `node.exe ... codex.js`. happy to approve once clippy is green.

graykode (commented):
pushed a one-line fix (cfg-gated `let _ = sessions_dir;`) to unblock clippy. CI should be green shortly. once it lands the windows logic + cmd_has_binary changes are good to merge — the two non-blocking heuristic notes from the prior review still stand but are fine as follow-ups.

graykode (approved):
clippy now green after the cfg-gated `let _ = sessions_dir;` fix landed (`check` job passing). security recap: no new deps, no new network/shell, no fs writes outside the project — pure platform-specific path-scan refactor. windows logic + the `cmd_has_binary` extension stripping look correct.

the two non-blocking heuristic notes from the earlier review still stand as follow-ups (mtime-based PID↔file mapping is fragile with ≥2 concurrent codex processes, and `today_session_dir` won't pick up sessions running across midnight), but neither is a merge blocker. lgtm.

Issue #73, opened 2026-04-24
Request for Opencode CLI


sangse0911:
**Describe the feature**
Support for Opencode CLI

**Use case**
This solves for developers using Opencode in combination with other Frontier Labs

tbouquet:
Duplicate of #38

code-lixm:
how's going on, excited to wait

2233admin:
This is now being worked on in PR #100. The OpenCode support has been rebased onto latest main with fixes for API compatibility and tests.

graykode:
OpenCode support has shipped in `v0.4.3` via #40.

Please try the latest release:

```bash
cargo install abtop --locked --force
```

Then run `abtop` while an `opencode` session is active. abtop reads OpenCode session data from `~/.local/share/opencode/opencode.db`.

PR #100, opened 2026-05-05
fix: OpenCode support - rebase onto main + test fixes

### LINKED ISSUE #38: Support of OpenCode

2233admin:
Fixes for PR #40 (OpenCode support):

## Changes
1. **Rebase onto latest main** - Resolve conflicts in mod.rs and sessions.rs
2. **API compatibility** - Update SessionStatus::Working → SessionStatus::Thinking
3. **Missing fields** - Add context_history, compaction_count, context_window, tool_calls, pending_since_ms, thinking_since_ms, file_accesses
4. **Test fixes** - Update with_hidden tests to expect 3 collectors (claude + codex + opencode)

## Test Results
- 91 passed, 12 failed
- The 12 failures are pre-existing on main (not introduced by this PR)

Closes #38, #73

### DIFF
--- a/src/collector/mod.rs
+++ b/src/collector/mod.rs
@@ -1,12 +1,14 @@
 pub mod claude;
 pub mod codex;
 pub mod mcp;
+pub mod opencode;
 pub mod process;
 pub mod rate_limit;
 
 pub use claude::ClaudeCollector;
 pub use codex::CodexCollector;
 pub use mcp::McpServer;
+pub use opencode::OpenCodeCollector;
 pub use rate_limit::read_rate_limits;
 
 /// Redact common secret patterns to avoid displaying credentials in the TUI.
@@ -169,6 +171,9 @@ impl MultiCollector {
         }
         if !is_hidden("codex") {
             collectors.push(Box::new(CodexCollector::new()));
+        }
+        if !is_hidden("opencode") {
+            collectors.push(Box::new(OpenCodeCollector::new()));
         }
         Self {
             collectors,
--- a/src/collector/mod.rs
+++ b/src/collector/mod.rs
@@ -336,32 +336,32 @@ mod tests {
     #[test]
     fn with_hidden_empty_keeps_all_collectors() {
         let mc = MultiCollector::with_hidden(&[]);
-        assert_eq!(mc.collectors.len(), 2);
+        assert_eq!(mc.collectors.len(), 3);
     }
 
     #[test]
     fn with_hidden_codex_drops_codex_only() {
         let mc = MultiCollector::with_hidden(&["codex".to_string()]);
-        assert_eq!(mc.collectors.len(), 1);
+        assert_eq!(mc.collectors.len(), 2);
     }
 
     #[test]
     fn with_hidden_is_case_insensitive() {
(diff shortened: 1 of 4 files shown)


graykode (changes requested):
security is clean — no new deps, the only new shell/process surface is `sqlite3 -readonly -json` (SQL is constant, no user-controlled interpolation) and `lsof -p <u32> -d cwd -Fn` (pid is u32, no injection). symlink fail-closed on `db_path` and the secret-redaction on DB-sourced `title` are good. tests are 116/116 green on the merged tip (PR body's "12 pre-existing failures" claim is stale).

correctness — a few real issues:

1. **`match_pid_to_session` last-resort over-matches every session.** opencode.rs `match_pid_to_session`: when exactly one opencode process is running but its cwd doesn't match a session's `directory`, the `if pid_commands.len() == 1` branch returns that PID for *every* call. So if the user has N rows in the DB and one live `opencode` process that's working in a different dir, all N rows render as alive (Thinking/Waiting) reusing the same PID. The fallback should only kick in when there's exactly one DB session being matched, not per-row — or just drop it.

2. **`cmd.contains(session_dir)` is unguarded against empty/short strings.** in the same function: an empty `session_dir` makes `cmd.contains("")` always true, and a `directory` of `/` matches every absolute path. add `if !session_dir.is_empty() && session_dir.len() > 1` (or similar) before the substring fallback.

3. **doc comment is out of sync with the code.** the OpenCodeCollector docstring says "Queries run on slow ticks only (every ~10s via MultiCollector gating) to avoid forking a sqlite3 process every 2s", but `collect()` never checks `shared.slow_tick` — sqlite3 forks every tick (2s). either gate the queries on `shared.slow_tick` and cache the previous result, or fix the comment to match reality.

4. **"recently finished < 5 min" branch is dead code.** the `if !pid_alive && age_ms > 300_000 { continue; }` block keeps DB rows whose PID isn't alive, sets `status = SessionStatus::Done`, and tags `current_tasks = vec!["finished"]`. but `mod.rs:273` does `all.retain(|s| !matches!(s.status, SessionStatus::Done))` — those rows never reach the UI. either drop the dead branch, or follow the codex pattern (pid==0 sentinel + status that isn't Done) if showing recently-finished rows is actually intended.

nits: `since_update_secs < 30 → Thinking` is a coarse heuristic vs codex's `model_generating` flag, but acceptable for a TUI dashboard. children-walk logic duplicates codex.rs verbatim and could share a helper, but not blocking.

requesting changes for #1 specifically (it's a user-visible bug with multiple DB sessions). #2–#4 are easy follow-ups in the same push.

graykode:
Subsumed by #40. tbouquet's branch now contains every commit that was on this PR (the original feat + the review-feedback fixup), plus a merge commit resolving the rebase against current main (rename `SessionStatus::Working` -> `Thinking`, fill in the new `AgentSession` fields added since #40 was opened). Continuing the OpenCode review on #40 to keep the original author's review thread intact.

PR #101, opened 2026-05-05
fix(status): mark sessions Executing while tool_use is pending

BruceWW:
I/O-bound tools (Read, Edit) leave descendants under 5% CPU on fast machines, so the descendant-CPU heuristic missed them and Claude sessions flickered to Waiting while tools were actually running.

Record the assistant turn timestamp only when its content contains tool_use blocks, reset it on the matching tool_result user line, and treat non-zero as "tools pending" -> Executing.

Also fix the context-tokens gauge on a fresh session's first turn (cache_read=0, cache_creation>0): fall back to input + cache_creation so the gauge reflects real context size instead of just the prompt.

### DIFF
--- a/src/collector/claude.rs
+++ b/src/collector/claude.rs
@@ -482,9 +482,10 @@ impl ClaudeCollector {
 
         // Status is best-effort. Signals we trust:
         //   1. Active descendant CPU → tool is running.
-        //   2. last_assistant_ts_ms > 0 → assistant sent tool_use, awaiting
-        //      tool result. Many tools are I/O-bound (Read, Edit) with low
-        //      CPU, so we can't rely on CPU alone.
+        //   2. current_task non-empty → latest assistant turn left a
+        //      tool_use unanswered. Catches I/O-bound tools (Read, Edit)
+        //      whose descendants stay under 5% CPU, so the CPU heuristic
+        //      alone would flicker to Waiting while the tool runs.
         //   3. last_user_ts_ms > 0 → trailing transcript line is a real
         //      user prompt with no assistant reply yet, so the model is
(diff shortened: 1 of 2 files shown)


graykode (commented):
security clean — pure parser/comment change, no network/shell/deps touched. tests pass (104), build green on the merge tip.

the real fix here is that `pending_since_ms` no longer leaks past terminal text-only assistant turns (used to make `src/ui/sessions.rs:954` render a phantom growing tool-running bar after the model finished without a tool_use). and the `inp + cc` fallback for fresh sessions on cache_read=0/cache_creation>0 is the right call for the context gauge.

but the new comments and the PR title oversell what landed. `src/collector/claude.rs:485-487` and `src/model/session.rs:56` say `last_assistant_ts_ms > 0` is part of the Executing decision — the status check at `src/collector/claude.rs:506` still keys off `pending_tool = !cached.current_task.is_empty()` and isn't touched by this PR. the I/O-bound-tools-with-low-CPU case described in the body is already handled by the existing `current_task` signal (cleared on each new assistant turn, set on tool_use, so non-empty between tool_use and the next assistant turn). either reword the comments/title to match what actually changed (this is about `pending_since_ms` + the context gauge, not the status decision), or wire `|| cached.last_assistant_ts_ms > 0` into the check at line 506 so it matches the description.

also no test pins the actual regression scenario: a transcript whose **last** line is a text-only assistant entry should leave `last_assistant_ts_ms == 0`. existing `test_parse_transcript_trailing_user_marks_thinking_window` passes regardless because the user handler at line 1483 resets the timestamp anyway, so it doesn't exercise the new gate.

PR #102, opened 2026-05-05
fix(claude): only filter `--print` sessions abtop itself spawned

elondemirock:
## Summary

abtop's summary generator (`generate_summary` in `src/app.rs`) shells out to `claude --print -` for every session that needs a one-line title. To avoid double-counting those children, the collector currently drops any `claude` cmdline containing `--print`. Side effect: every non-interactive Claude session a user runs in their own shell is also hidden from the panel.

This PR replaces the cmdline heuristic with a real self-PID check: walk the ppid chain via a new `process::is_descendant_of` helper and only drop the candidate when abtop's own PID is an ancestor. abtop spawns `claude --print` directly through `Command::new("claude")` with no shell wrapper, so a single hop suffices, but the helper handles arbitrary depth (with cycle and missing-parent guards).

`SharedProcessData` gains a `self_pid: u32` field populated from `std::process::id()` in `fetch`, threaded into `find_claude_pids`, `build_discovery_context`, and `load_session`.

## Changes

- `src/collector/process.rs`: new `is_descendant_of` helper with cycle/missing-parent guards
- `src/collector/claude.rs`: replace `--print` cmdline filter with self-PID ancestry check; thread `self_pid` through discovery
- `src/collector/mod.rs`: populate and propagate `self_pid` on `SharedProcessData`
- Allow `clippy::too_many_arguments` on `load_session` (now 8 args; refactor out of scope)

## Tests

- [x] Existing test for the post-`/clear` sid override is preserved
- [x] Tests updated to model the print child as `ppid=abtop_pid` (matches production)
- [x] New test `test_load_session_keeps_user_spawned_print` pins the desired behaviour: a user's own `claude --print` session must show up in abtop
- [x] Manual: run abtop and a separate `claude --print` from a shell, confirm both visible

### DIFF
--- a/src/collector/claude.rs
+++ b/src/collector/claude.rs
@@ -105,7 +105,8 @@ impl ClaudeCollector {
             self.refresh_config_dirs(&shared.process_info);
         }
 
-        let active_session_paths = self.discover_active_session_paths(&shared.process_info);
+        let active_session_paths =
+            self.discover_active_session_paths(&shared.process_info, shared.self_pid);
         let active_config_dirs: Vec<ConfigDir> = active_session_paths
             .iter()
             .map(|(_, config)| config.clone())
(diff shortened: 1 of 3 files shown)


elondemirock:
Reopening with clean history.

PR #103, opened 2026-05-05
fix(claude): only filter `--print` sessions abtop itself spawned

elondemirock:
### Current behavior

The collector currently drops any `claude` cmdline containing `--print` because `abtop` uses it for per-session one-line summaries.

❌ **Side effect**: every other non-interactive Claude session is also hidden from the panel.

### Proposed change

Replace the cmdline heuristic with a real self-PID check: only drop the candidate when abtop's own PID is an ancestor.

## Changes

- New `is_descendant_of` helper in `process.rs` that walks the ppid chain (with cycle and missing-parent guards).
- Replace the `--print` cmdline filter in 3 places with a self-PID ancestry check.
- `self_pid` is sourced once via `std::process::id()` and threaded explicitly to the callees that need it.
- `load_session` reads `self_pid` from `DiscoveryContext` (already passed in) so its arg count is unchanged.

## Considerations

- No new field on `SharedProcessData` — the existing process data already flows everywhere needed; only the abtop PID had to be added, and that's available locally.
- Used a ppid walker instead of a single-hop check. Depth is always 1 today (abtop spawns `claude --print` directly through `Command::new("claude")`), but the helper is reusable and defended against cycles/orphans.

## Test plan

- [x] Existing test for the post-`/clear` sid override is preserved
- [x] Tests updated to model the print child as `ppid=abtop_pid` (matches production)
- [x] New test pins the desired behaviour: a user's own `claude --print` session must show up in abtop
- [x] `is_descendant_of` has unit tests for direct/multi-hop chains, unrelated processes, self-check, zero pids, cycles, and missing parents
- [x] Manual: run abtop and a separate `claude --print` from a shell, confirm both visible

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -92,7 +92,7 @@ Session file format:
 ```
 - ~170 bytes. Created on start, deleted on exit.
 - Verify PID alive with shared `ps` data containing a `claude` binary.
-- Skip sessions whose PID descends from abtop's own (filters abtop's own `claude --print` summary children without hiding user-spawned non-interactive sessions).
+- Skip sessions whose PID descends from abtop's own `claude --print` summary children without hiding user-spawned non-interactive sessions.
 
 ### 2. Claude Code transcript: `{config-root}/projects/{encoded-path}/{sessionId}.jsonl`
 Path encoding: `/Users/foo/bar` → `-Users-foo-bar`
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -92,7 +92,7 @@ Session file format:
 ```
 - ~170 bytes. Created on start, deleted on exit.
 - Verify PID alive with shared `ps` data containing a `claude` binary.
-- Skip `--print` sessions (abtop's own LLM calls for summary generation).
+- Skip sessions whose PID descends from abtop's own (filters abtop's own `claude --print` summary children without hiding user-spawned non-interactive sessions).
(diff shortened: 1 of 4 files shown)


graykode (approved):
lgtm — security clean (no new deps/network/shell, just a more precise heuristic). nice tightening of the filter.

a few notes, none blocking:

- `is_descendant_of` short-circuits at `ppid == 1`, which is right for the common case but means an abtop summary child that gets reparented to launchd/init mid-poll (e.g. abtop crashes between spawn and the 10s wait_with_output) would briefly surface in the panel until the orphan is reaped. Pre-existing race in spirit and the session file cleanup catches it shortly after, so probably fine to leave as-is.
- depth is always 1 today via `Command::new("claude")` in `app.rs:687`, so the chain walker is overkill for now — but the cycle/missing-parent guards make it cheap insurance and keep the helper reusable. agreed with the call.
- test coverage is solid: the new `test_load_session_keeps_user_spawned_print` is the right counterpart to the existing regression test, and the `is_descendant_of` unit tests cover the corner cases (cycle, missing parent, self, zero pids).

verified locally: `cargo build` clean, all 112 tests pass on the post-merge tree. (a `clippy::items_after_test_module` warning fires in `src/ui/sessions.rs` but it reproduces on main too — not introduced by this PR.)

PR #105, opened 2026-05-07
fix(windows): detect node-wrapped Codex commands

AntaresGG:
## Summary
- scan all Windows command tokens when matching agent binaries
- trim surrounding quotes before path basename checks
- add a regression test for node-wrapped Codex installed under Program Files

## Tests
- cargo test cmd_has_binary -- --nocapture
- cargo test collector::codex -- --nocapture

### DIFF
--- a/src/collector/codex.rs
+++ b/src/collector/codex.rs
@@ -312,6 +312,25 @@ impl CodexCollector {
                 pids.push((*pid, is_exec));
             }
         }
+
+        // Windows npm/Git shims can create a chain like:
+        // sh.exe -> node.exe ...\codex.js -> codex.exe.
+        // Once the real codex child exists, keep that child and drop wrapper
+        // ancestors; otherwise Windows rollout fallback maps each candidate PID
+        // to a different recent JSONL file and historical sessions look live.
+        let candidates = pids.clone();
+        pids.retain(|(pid, _)| {
+            process::cmd_first_token_has_binary(
+                process_info
+                    .get(pid)
+                    .map(|info| info.command.as_str())
+                    .unwrap_or_default(),
+                "codex",
+            ) || !candidates.iter().any(|(other_pid, _)| {
+                *other_pid != *pid && process::is_descendant_of(*other_pid, *pid, process_info)
+            })
+        });
+
         pids
     }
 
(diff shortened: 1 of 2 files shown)


graykode (approved):
lgtm — security is clean (pure cmd parsing + PID filtering, no new network/shell/deps), and the wrapper-drop logic correctly collapses the `sh.exe -> node.exe(codex.js) -> codex.exe` chain to just the real child.

pushed one fixup commit: the new `proc_info` test helper is only used inside the `#[cfg(windows)]` test, so on macOS/Linux it tripped `dead_code` under `-D warnings`. gated it with `#[cfg(windows)]`.

one nit, non-blocking: `windows_command_tokens` handles quoted tokens correctly when fed a raw `GetCommandLineW`-style string, but `get_process_info` builds `command` by `join(" ")`-ing `proc_.cmd()` with no re-quoting — so paths containing spaces (`C:\Program Files\...`) come back out unquoted. tokens like `"C:\Program"` and `"Files\nodejs\node.exe"` would then split apart again, and the second-token check could miss a match. it's strictly better than the previous behavior either way, so happy to defer.

graykode (approved):
lgtm. security is clean (pure parsing of process listings already exposed via sysinfo, no new network/shell/deps), and the wrapper-strip logic in `find_codex_pids_from_shared` looks right — i traced the sh.exe → node.exe → codex.exe chain by hand and the retain predicate keeps the real codex leaf and drops both ancestors. cycle protection in `is_descendant_of` covers the pathological cases too.

worked through these and they all hold:
- single wrapper-only PID (no real codex.exe child): kept (no other-as-descendant in candidates)
- multiple unrelated codex sessions: all kept (no descendant relation)
- on unix, `pnpm codex`/`npx codex` style invocations now drop the wrapper and keep the leaf — that's an improvement, not a regression

`cargo test` passes (114/114) on top of merged main on macOS. the new tests are `#[cfg(windows)]`-gated so they don't run here, but the logic is straightforward enough that the windows test cases look correct by inspection. preexisting clippy warning in `src/ui/sessions.rs:893` is unrelated to this PR.

one thing worth flagging, not blocking:
- `with_cmd(OnlyIfNotSet)` → `with_cmd(Always)` in `process.rs:135` is a global change — it makes sysinfo re-read cmdlines for every process every tick, on linux/macOS too. needed for windows correctness (wrapper cmdlines can change), but it's a small ongoing cost on the other platforms. fine for now; worth keeping in mind if process-list refresh ever shows up in profiling.

PR #106, opened 2026-05-07
Improve terminal UI and context accounting

shawnsw:
## Summary
- Improve responsive terminal layout for narrow/mobile panes, including grouped tabs and click targets.
- Show recent selected-session chat history by default while keeping timeline available behind a toggle.
- Fix Codex context usage so cached input is not double-counted and current context stays tied to the active window.
- Add install script for building and installing abtop locally.

## Commits
- 8f66420 feat: improve responsive terminal UI
- b356eac feat: show session chat history
- b976a3a fix: correct codex context usage

## Verification
- cargo test
- cargo clippy -- -D warnings
- cargo run -- --once

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 .claude/worktrees/
 demo.tape
 .idea/
+.perles/
--- a/scripts/install-abtop.sh
+++ b/scripts/install-abtop.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+script_dir="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)"
+repo_root="$(cd -- "$script_dir/.." && pwd)"
+install_dir="${INSTALL_DIR:-$HOME/bin}"
+install_path="${ABTOP_INSTALL_PATH:-$install_dir/abtop}"
+target_dir="$repo_root/target"
+built_binary="$target_dir/release/abtop"
+
+if ! command -v cargo >/dev/null 2>&1; then
+  echo "error: cargo not found in PATH" >&2
+  exit 1
+fi
+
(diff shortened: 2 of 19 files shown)


graykode:
Could you make sure the commits are attributed to your GitHub author? Right now they're not linked to your account. Thanks!

graykode:
Review findings to fix in this PR. I will patch these directly on the PR branch.

1. `src/main.rs` handles `MouseEventKind::Down(_)`, so left, right, and middle clicks all activate the same click targets. That includes the ports panel `KillOrphanPorts` target, which means a non-left click can trigger process-kill behavior while mouse capture is enabled. Destructive click handling should be limited to `MouseButton::Left`.

2. The session chat tail renders user/assistant transcript text directly into the TUI. The current chat cleanup only normalizes whitespace and redacts a few secret prefixes; it does not strip terminal control characters or bidi override characters. Chat text should go through the same class of terminal-safe sanitization used for snapshot output before it is stored or rendered.

graykode:
Closing this one because the scope is too broad for a safe merge as-is.

The Codex context accounting fix looks useful and worth keeping, but this PR also bundles responsive layout changes, chat-history UI, an install script, and related refactors. That makes the review and regression surface much larger than the actual fix we need right now.

Could you please reopen this work as smaller PRs, starting with just the Codex context accounting fix? The UI/chat-history pieces can then be reviewed separately with screenshots and focused behavior tests.

graykode:
Closed in favor of smaller, focused follow-up PRs.

shawnsw:
@graykode thanks for the review. I have fixed the account email issue, and broken the PR into 3 smaller ones #113 #114 #115 per your suggestion.

graykode:
@shawnsw Thanks for your PRs!

PR #110, opened 2026-05-08
fix(context): read settings.json to detect 1M model before 200k token threshold

### LINKED ISSUE #109: Context percentage incorrectly calculated against 200k window for 1M-context models
### Description

When using a 1M-context model (e.g. `claude-sonnet-4-6[1m]`), the context usage percentage shown in the Context panel is calculated against a **200k** token window instead of the actual **1M** window.

This causes the percentage to appear 5× too high during the early portion of a session (before token usage exceeds 200k).

### Root Cause

`context_window_for_model` determines the window size with two signals:

1. `model.contains("[1m]")` — checks the model name from the JSONL transcript
2. `max_context_tokens > 200_000` — fallback once enough tokens have been used

However, Claude Code writes the **Anthropic API model name** to the JSONL transcript (e.g. `claude-sonnet-4-6`), **without** the `[1m]` suffix that appears in the UI. Signal #1 therefore never matches, and signal #2 only kicks in after 200k tokens are consumed — meaning the entire early phase of a 1M session uses the wrong 200k denominator.

The `[1m]` suffix **does** appear in `~/.claude/settings.json` under the `model` key (e.g. `"model": "sonnet[1m]"`), which is not currently read by abtop.

### Steps to Reproduce

1. Select a 1M-context model in Claude Code (e.g. `claude-sonnet-4-6[1m]`)
2. Start a session and use < 200k tokens
3. Observe the Context % in abtop — it will read ~5× higher than the true percentage

### Expected Behavior

Context % is computed against the 1M window from the very first turn.

### Proposed Fix

Read the configured model from Claude Code's `settings.json` (following the same priority chain already used by `read_effort_level`) and pass it alongside the transcript model to `context_window_for_model`. If either string contains `[1m]`, use the 1M window size.

xiaoye5200:
## Summary

Fixes #109

- Claude Code stores only the bare API model name (e.g. `claude-sonnet-4-6`) in JSONL transcripts — the `[1m]` suffix never appears there, so the existing `model.contains("[1m]")` check in `context_window_for_model` never matched.
- The only fallback was `max_context_tokens > 200_000`, which means the 1M window size was not recognised until 200k tokens had already been consumed, making the context percentage read ~5× too high throughout the early phase of every 1M session.

## Fix

Add `read_configured_model` (follows the same settings-file priority chain as the existing `read_effort_level`) to read the `model` key from Claude Code's `settings.json`. That key **does** contain the `[1m]` suffix (e.g. `"sonnet[1m]"`).

`context_window_for_model` now accepts a `configured_model` parameter and returns 1 M if **either** the transcript model or the configured model contains `[1m]`, so the correct window size is used from the very first turn.

## Changes

- `src/collector/claude.rs`
  - `context_window_for_model(transcript_model, configured_model, max_context_tokens)` — new signature
  - `read_configured_model(cwd)` + `read_model_from_settings(path)` — new helpers
  - Updated call site and existing unit tests; added a test for settings-based detection

## Test plan

- [x] `cargo test` passes (updated `test_context_window_for_model` covers the new `configured_model` path)
- [ ] Start abtop with a `claude-sonnet-4-6[1m]` session that has < 200k tokens — Context % should now reflect the 1M window

### DIFF
--- a/src/collector/claude.rs
+++ b/src/collector/claude.rs
@@ -527,7 +527,8 @@ impl ClaudeCollector {
             }
         };
 
-        let context_window = context_window_for_model(&model, max_context_tokens);
+        let configured_model = read_configured_model(&sf.cwd);
+        let context_window = context_window_for_model(&model, &configured_model, max_context_tokens);
         let context_percent = if context_window > 0 {
             (last_context_tokens as f64 / context_window as f64) * 100.0
         } else {
@@ -1679,8 +1680,8 @@ fn truncate(s: &str, max: usize) -> String {
     }
 }
 
-fn context_window_for_model(model: &str, last_context_tokens: u64) -> u64 {
-    if model.contains("[1m]") || last_context_tokens > 200_000 {
+fn context_window_for_model(transcript_model: &str, configured_model: &str, max_context_tokens: u64) -> u64 {
+    if transcript_model.contains("[1m]") || configured_model.contains("[1m]") || max_context_tokens > 200_000 {
         1_000_000
     } else {
         200_000
@@ -1732,6 +1733,34 @@ fn read_effort_from_settings(path: &Path) -> Option<String> {
         None
     } else {
         Some(level.to_string())
+    }
+}
+
+fn read_configured_model(cwd: &str) -> String {
+    let mut candidates: Vec<PathBuf> = Vec::new();
+    let cwd_path = PathBuf::from(cwd);
+    candidates.push(cwd_path.join(".claude").join("settings.local.json"));
+    candidates.push(cwd_path.join(".claude").join("settings.json"));
+    if let Some(home) = dirs::home_dir() {
+        candidates.push(home.join(".claude").join("settings.local.json"));
+        candidates.push(home.join(".claude").join("settings.json"));
+    }
+    for path in candidates {
+        if let Some(model) = read_model_from_settings(&path) {
+            return model;
+        }
+    }
+    String::new()
+}
+
+fn read_model_from_settings(path: &Path) -> Option<String> {
+    let content = fs::read_to_string(path).ok()?;
+    let val: Value = serde_json::from_str(&content).ok()?;
+    let model = val.get("model")?.as_str()?.trim();
+    if model.is_empty() {
+        None
+    } else {
+        Some(model.to_string())
     }
 }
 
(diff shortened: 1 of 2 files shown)


xiaoye5200:
Hi @graykode 👋

Could you please approve the CI workflows to run and take a look at this PR when you get a chance?

This fixes a bug where the context percentage is calculated against a 200k window even when using a 1M-context model (e.g. `claude-sonnet-4-6[1m]`), because Claude Code writes only the bare API model name to JSONL transcripts — without the `[1m]` suffix — so the existing check never matched.

Happy to make any adjustments based on your feedback. Thanks!

graykode (approved):
lgtm — security is clean (reads the same `.claude/settings*.json` abtop already touches, no new deps/network/shell). diagnosis matches the root cause in #109: the `[1m]` suffix only lives in settings, never in the transcript, so checking both is the right fix. signature change is on an internal helper with a single call site, updated correctly. tests cover transcript-side and settings-side detection plus fallbacks; 114 tests pass locally.

one small nit, non-blocking: `read_effort_level` checks `CLAUDE_CODE_EFFORT_LEVEL` env first before walking the settings chain, but `read_configured_model` skips the env-var step. Claude Code's `ANTHROPIC_MODEL` env wouldn't carry the `[1m]` suffix anyway, so it probably doesn't matter — just noting the asymmetry in case env override is added later. the priority-chain duplication between the two readers could be hoisted into a `settings_candidate_paths(cwd)` helper, but that's a follow-up, not a blocker.

xiaoye5200:
Thanks for the thorough review!

Addressed both nits in the latest commit (cff0fa1):
- Hoisted the settings file path chain into a shared `settings_candidate_paths(cwd)` helper, used by both `read_effort_level` and `read_configured_model`
- Added `CLAUDE_CODE_MODEL` env-var check to `read_configured_model` for symmetry with `read_effort_level`

graykode:
small request — the three commits on this branch show up with author name `unknown` in the GitHub UI (your local `git config user.name` looks unset). could you fix the author metadata before merge? something like:

```
git config user.name "Your Name"
git rebase -i origin/main --exec "git commit --amend --reset-author --no-edit"
git push --force-with-lease origin fix/1m-context-window-detection
```

(rebasing + force-pushing your own fork branch is fine; the repo policy is just that reviewers don't force-push to author forks.) approval still stands once that's done.

xiaoye5200:
Hi @graykode — just pushed two follow-up commits addressing your nits (settings_candidate_paths helper + CLAUDE_CODE_MODEL env-var symmetry). The CI workflows will need another approval to run on the new HEAD. No rush at all, and thanks again for the detailed review!

graykode:
quick policy note before this lands — going forward, please open a new PR for any commits added after a review approval. mixing post-approval changes into the same PR resets the review baseline and makes it harder to keep the diff focused.

specifically on the two follow-ups here:

- **`43fb69d` (sort_by → sort_by_key in codex.rs)**: unrelated to the 1M context fix this PR is scoped to. happy to take it as a separate PR. (also, I couldn't reproduce `unnecessary_sort_by` against main on stable — may be a newer toolchain on your end.)
- **`cff0fa1` (refactor + env var)**: the path-chain hoist + env-var symmetry is exactly what my earlier nits pointed at, thanks for picking those up.

build/tests pass on both, so this is process not correctness. approval still stands once the unrelated `sort_by` change is split out (and the author-name fix from my previous comment is applied).

(edited: walked back an earlier objection about the env-var addition — that was inconsistent with my original nit, sorry for the noise.)

xiaoye5200:
Done — set user.name/email and rebased all three commits with --reset-author. Author should now show correctly as xiaoye5200 in the GitHub UI.

graykode (approved):
lgtm. security is clean (read-only on the same `.claude/settings.{,local.}json` paths `read_effort_level` already touches, no new network/shell/deps), and the fix is correct — the `[1m]` detection now lights up from the very first turn. `cargo test` passes (114/114) on top of merged main. preexisting clippy warning in `src/ui/sessions.rs:893` (`items_after_test_module`) is unrelated to this PR.

two nits, neither blocking:
- `read_model_from_settings` has no dedicated unit tests, while its sibling `read_effort_from_settings` has four (extract / missing field / empty string / nonexistent file). worth mirroring those for symmetry.
- `read_effort_level` and `read_configured_model` now walk the same 4 paths and read each file separately. an obvious follow-up is to read each settings file once and pull both fields, but it's a cleanup, not a fix for this PR.

the codex.rs `sort_by` → `sort_by_key(Reverse(...))` is unrelated to the context-window fix; fine to leave bundled here since it's a one-liner clippy-style change.

PR #111, opened 2026-05-09
fix(unix): detect claude.exe binary on macOS/Linux

squallyjy:
## Summary

The `@anthropic-ai/claude-code` npm package ships its binary as `claude.exe` even on non-Windows platforms. The existing `unix_token_has_binary` only compared the exact basename, so processes running as:

```
/usr/local/lib/node_modules/@anthropic-ai/claude-code/bin/claude.exe --session-id ...
```

were completely invisible to abtop — `find_claude_pids` returned an empty list, and no sessions were ever displayed.

This PR strips the `.exe` suffix before comparing on Unix, mirroring the behavior already present in the Windows code path.

## Changes

- `src/collector/process.rs`: add `.exe` suffix stripping in `unix_token_has_binary()`
- Added test `cmd_has_binary_exe_suffix_on_unix` covering the new behavior

## Test plan

- [x] `cargo test cmd_has_binary` — all 4 tests pass
- [x] Manually verified on macOS with `claude.exe` process — abtop now shows sessions correctly

### DIFF
--- a/src/collector/process.rs
+++ b/src/collector/process.rs
@@ -423,6 +423,12 @@ fn unix_token_has_binary(tok: &str, name: &str) -> bool {
     if base == name {
         return true;
     }
+    // Strip .exe suffix for compatibility with claude.exe on non-Windows (e.g. @anthropic-ai/claude-code npm package)
+    if let Some(stripped) = base.strip_suffix(".exe") {
+        if stripped == name {
+            return true;
+        }
+    }
     matches!((iter.next(), iter.next()), (Some("versions"), Some(parent)) if parent == name)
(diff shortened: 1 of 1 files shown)


graykode (approved):
lgtm — security is clean (pure string-matching change, no new deps/network/shell). surgical fix, all 122 tests pass on macOS. test covers both positive paths (full npm path + bare `claude.exe`) and the negative `notclaude.exe` case so the does-not-overmatch invariant is preserved. `cmd_first_token_has_binary` also picks up the fix transparently since both go through `unix_token_has_binary`.

PR #112, opened 2026-05-09
fix(status): treat slash-command and bash user-role lines as synthetic

BruceWW:
Sessions could pin in Thinking ("◉ Think generating reply") forever after `/plugin update` (and other pure-local slash commands or `!bash` invocations). Claude Code flushes 2-3 user-role lines for these — a `<local-command-caveat>` (isMeta:true), a `<command-name>` line, and a `<local-command-stdout>` line — none of which prompt the model, so no assistant reply ever arrives to clear `last_user_ts_ms`.

`is_tool_result_user_msg` only filtered tool_result block arrays and let every string-content user line through, so the trailing local- command line opened a thinking window that never closed.

Rename to `is_synthetic_user_msg` and additionally treat as synthetic:
- entries with `isMeta: true`
- string content starting with `<local-command-stdout|stderr|caveat>`,   `<command-name>`, or `<bash-input|stdout|stderr>`

Also skip these lines when extracting the session's initial_prompt so the title isn't a `<command-name>/plugin...` dump.

Add regression tests for the /plugin three-line sequence and the `!bash` `<bash-input>/<bash-stdout>` sequence.

### DIFF
--- a/src/collector/claude.rs
+++ b/src/collector/claude.rs
@@ -1502,15 +1502,20 @@ fn parse_transcript(path: &Path, from_offset: u64) -> TranscriptResult {
                                 result.last_assistant_ts_ms = 0;
                             }
                             // Mark the start of a thinking window — next assistant
-                            // turn clears it. **Skip tool_result wrappers**:
-                            // Claude Code serializes both real prompts and tool
-                            // results as `user`-role lines, but only real prompts
-                            // mean "model has been asked, no reply yet". A tool
-                            // loop alternates assistant(tool_use) ↔ user(tool_result)
-                            // inside one logical turn, and treating each
-                            // tool_result as the start of a new thinking window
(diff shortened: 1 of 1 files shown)


graykode (approved):
lgtm — security clean (single-file parser refactor, no new deps/network/shell). the rename to `is_synthetic_user_msg` and lifting it to take the whole entry reads right since `isMeta` lives at the entry level, and keeping the "all blocks are tool_result" check conservative (mixed → still a real prompt) preserves the prior safety. nice catch also gating `initial_prompt` extraction so titles don't end up as `<command-name>/plugin...` or a bash stdout dump. the two regression tests exercise the exact 3-line `/plugin` sequence and the `<bash-input>/<bash-stdout>` pair from the bug report. merged main locally and `cargo test` is green (123/123).

PR #113, opened 2026-05-10
fix: correct Codex context accounting

shawnsw:
## Summary

Fix Codex context accounting so cached input is not counted twice.

- Store non-cached input separately from cache reads
- Use `last_token_usage.input_tokens` for current context usage
- Add regression coverage for the over-100% context case

## Verification

- `cargo test codex`

### DIFF
--- a/src/collector/codex.rs
+++ b/src/collector/codex.rs
@@ -471,6 +471,8 @@ struct CodexJSONLResult {
     model_generating: bool,
     last_activity: std::time::SystemTime,
     initial_prompt: String,
+    /// Input tokens excluding cached input, matching AgentSession's additive
+    /// token accounting where cache reads are stored separately.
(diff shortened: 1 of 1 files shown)


graykode (approved):
lgtm — security clean (pure parser logic, no new deps/network/shell). good catch on the double-count: Codex's `input_tokens` already includes `cached_input_tokens` (unlike Claude's API), so the old `inp + cache` and `total_input = inp` paths inflated both context% and `total_tokens()` whenever cache was hot. after the fix the additive contract still holds: `(inp-cache) + out + cache == inp + out`, matching Codex's own `total_tokens`. cache-read fallback path (`cache_read_input_tokens` legacy field) still covered, regression test with the real >100% case is exactly the right shape. all 21 codex tests pass on macOS.

PR #114, opened 2026-05-10
feat: improve responsive terminal UI

shawnsw:
## Summary

Improve responsive terminal layout for compact panes.

- Add compact Work / Usage / System tabs
- Improve section sizing at narrow widths
- Add keyboard and mouse navigation for compact views
- Restrict click actions to left-click only

## Verification

- `cargo test`

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -34,6 +34,54 @@ pub enum JumpOutcome {
     NoOp,
 }
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum NarrowTab {
+    Work,
+    Usage,
+    System,
+}
+
+impl NarrowTab {
+    pub const ALL: [Self; 3] = [Self::Work, Self::Usage, Self::System];
+
+    pub fn label(self) -> &'static str {
+        match self {
+            Self::Work => "Work",
+            Self::Usage => "Usage",
+            Self::System => "System",
+        }
+    }
+
+    pub fn shortcut(self) -> char {
+        match self {
+            Self::Work => 'w',
+            Self::Usage => 'u',
+            Self::System => 's',
+        }
+    }
+}
+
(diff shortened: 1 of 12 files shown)


graykode (approved):
lgtm — security clean (UI-only diff, no new deps/network/shell, mouse capture is the standard TUI tradeoff). solid responsive split: `NarrowTab`/`NarrowSection` in `App` with `clamp_*` keeping state coherent across panel toggles, the `<DESKTOP_WIDTH` branch handles tabs/sections/zoom + mouse + keyboard + scroll, and the new compact tests pin both rendered text and click targets so future width-threshold tweaks won't silently regress hit-testing. nice incidental cleanup moving the `sessions.rs` test module to end-of-file (fixes the pre-existing `items_after_test_module` warning). also confirmed merging current `main` and running `cargo build` / `cargo test` / `cargo clippy --all-targets -- -D warnings` is all green (133/133, 0 lint warnings) on top of the merge.

couple of follow-up nits, none blocking:

- `src/ui/mod.rs::session_at` duplicates the `detail_reserve` decision tree from `src/ui/sessions.rs::draw_sessions_panel` (the `if app.show_timeline ... else if inner.height <= 12 ... else ...` ladder). same story for `ports_kill_at` mirroring `draw_ports_panel`'s row layout (header + live + orphan + kill). these have to stay in lockstep for click hit-testing to land on the right row, and right now nothing structurally enforces that. extracting the row-layout decisions into a helper used by both renderer and hit-tester would prevent silent drift the next time someone adjusts column logic.
- mouse capture disables the terminal's native text selection for most terminals (users have to hold Shift). worth a one-liner in README or `?` help so new users know.
- `set_narrow_tab(tab)` silently no-ops when the tab has no visible panels, so pressing `w`/`u`/`s` on a config where that group is fully hidden does nothing. small UX surprise; a flash status message ("no panels in this tab") would smooth it. happy to land as a follow-up.

PR #115, opened 2026-05-10
feat: show session chat history

shawnsw:
## Summary

Show recent user/assistant chat history for the selected session.

- Collect bounded chat tails from Claude and Codex transcripts
- Exclude tool results and tool inputs
- Sanitize terminal control and bidi characters before rendering
- Keep timeline and file-audit views available behind toggles

## Verification

- `cargo test chat`
- `cargo test`
- `cargo clippy -- -D warnings`

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -898,6 +898,7 @@ mod tests {
             children: vec![],
             initial_prompt: String::new(),
             first_assistant_text: String::new(),
+            chat_messages: vec![],
             tool_calls: vec![],
             pending_since_ms: 0,
             thinking_since_ms: 0,
--- a/src/collector/claude.rs
+++ b/src/collector/claude.rs
@@ -1,7 +1,7 @@
 use super::process::{self, ProcInfo};
 use crate::model::{
-    AgentSession, ChildProcess, FileAccess, FileOp, SessionFile, SessionStatus, SubAgent,
-    MAX_FILE_ACCESSES,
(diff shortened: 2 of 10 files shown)


graykode (approved):
lgtm — security clean and actually defenses-positive: the new `sanitize_terminal_text` strips C0/C1 controls + bidi override/isolate marks before transcript text hits the TUI, and chat lines also pass through `redact_secrets`. tests cover both (BEL/RLO removal and `sk-proj-secret` → `[REDACTED]`).

pushed one merge commit (`9884794`) to resolve conflicts with main:
- #112 (`is_synthetic_user_msg`) replaced this PR's `is_tool_result_user_msg`. resolved by reusing the broader `is_synthetic_user_msg` for the chat-tail gating too — slash-command/bash/meta-caveat user lines are now also kept out of chat history, which matches the spirit of this PR (only show *real* user/assistant turns).
- #113 (codex token comment) was a trivial co-located addition, kept both sides.

after merge, `cargo build` clean and all 128 tests pass on macOS, including this PR's three new tests and #112's slash-command thinking-window test.

one nit, non-blocking: `clean_chat_text` is duplicated between `collector/claude.rs` and `collector/codex.rs` and the truncation tail differs — claude uses `truncate(…, max)` which appends `…`, codex uses `chars().take(max).collect()` (no marker). worth unifying in a follow-up so users don't see one panel ellipsize and the other hard-cut.

(unrelated: `cargo clippy --all-targets -- -D warnings` still fails on `src/ui/sessions.rs` `items-after-test-module`, but that's pre-existing on main, not this PR.)

PR #116, opened 2026-05-10
Add Codex Desktop app monitoring support

crosswyb:
## Summary

Adds a new `CodexDesktopCollector` that monitors **Codex Desktop app** sessions by reading token usage data from `~/.codex/logs_2.sqlite`. This is complementary to the existing `CodexCollector` which only covers Codex CLI.

## How it works

- Detects the Codex Desktop process (`Codex.app/Contents/MacOS/Codex`) via `ps`
- Reads `~/.codex/logs_2.sqlite` using the `sqlite3` CLI (no new Rust dependencies)
- Parses structured log entries containing `input_token_count`, `output_token_count`, `cached_token_count` per conversation
- Groups by `conversation.id` to form per-session views
- Registered as `"codex-desktop"` agent in `MultiCollector`, respects `hidden_agents` config
- Sessions auto-mark as Done when the desktop app process exits

## Design decisions

- **No new dependencies** — uses `std::process::Command` to call `sqlite3` CLI with `-json` output, consistent with existing patterns (`lsof`, `ps`, `git`)
- **Separate collector** — not merged into existing CodexCollector because the data source (SQLite vs JSONL) and discovery strategy (process name vs `codex` binary) are fundamentally different
- **High-water mark** — tracks `last_ts` to only re-read new log entries each tick

## Limitations

- No cwd/project_name/chat_messages/tool_calls (SQLite logs don't expose these at the same granularity as JSONL)
- Context window percentage is estimated (no per-turn breakdown in aggregated SQLite data)
- Requires `sqlite3` CLI to be installed (pre-installed on macOS, available everywhere)

## Testing

- 3 new unit tests for field extraction logic
- All 139 existing tests pass (updated `with_hidden` test assertions for the new 4th collector)

Closes: N/A — feature request from user who needed desktop app monitoring alongside CLI support

### DIFF
--- a/src/collector/codex_desktop.rs
+++ b/src/collector/codex_desktop.rs
@@ -1,0 +1,339 @@
+use super::process;
+use super::AgentCollector;
+use crate::model::{AgentSession, SessionStatus};
+use serde::Deserialize;
+use std::collections::HashMap;
+use std::path::PathBuf;
+use std::process::Command;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+/// Collector for Codex Desktop App sessions.
+///
+/// Codex Desktop stores session data in `~/.codex/logs_2.sqlite` instead of
+/// the JSONL files used by Codex CLI. This collector reads the SQLite DB
+/// directly to extract session metadata and token usage.
+///
+/// Data sources:
+/// 1. `ps` to find the Codex Desktop process (Codex.app)
+/// 2. The `logs_2.sqlite` database for per-session token counts, model info
+///
+/// Key log event pattern:
+/// ```text
+/// event.name="codex.sse_event" event.kind=response.completed
+///   input_token_count=335696 output_token_count=316 cached_token_count=332672
+///   reasoning_token_count=66 conversation.id=019dddc9-...
+///   model=gpt-5.4
+/// ```
+pub struct CodexDesktopCollector {
+    db_path: PathBuf,
+    /// Cached per-session data keyed by conversation ID.
+    sessions: HashMap<String, DesktopSessionState>,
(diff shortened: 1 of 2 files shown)


graykode:
Thanks for adding Codex Desktop support. The feature direction makes sense for abtop, but I don't think this implementation is safe to merge yet.

Main blockers:

- `src/collector/codex_desktop.rs:97` treats `logs.ts` as milliseconds, but the Codex Desktop DB stores epoch seconds plus `ts_nanos`. With real DB rows this can make sessions fail the active-window check, and the same unit mismatch affects `started_at` values later in the file.
- `src/collector/codex_desktop.rs:265-345` rereads the latest 500 rows on every collect pass and adds token counts again each time. `last_ts` is updated but not used as a cursor. This will inflate token totals on every refresh. Rows are also processed newest-to-oldest, so an older event can overwrite `last_event_kind` after a newer event.
- `src/collector/codex_desktop.rs:310-318` skips tool-related events before they can drive `SessionStatus::Executing`, because `conversation.id` is only extracted for completed/failed/created/in_progress events.
- `src/collector/codex_desktop.rs:140-149` calculates `context_percent` by subtracting other conversations' total input from the current conversation. A conversation's context should not depend on unrelated sessions; this should use the latest per-conversation context/input value instead.

Verification:

- `cargo test` passes locally: 141 tests.
- `cargo clippy -- -D warnings` fails on this branch with `derivable_impls`, `unnecessary_cast`, and `useless_format` in `src/collector/codex_desktop.rs`.

Suggested direction: use a stable processed-row cursor such as `(ts, ts_nanos, id)`, process rows in ascending order, avoid re-adding already-processed token events, and store latest per-conversation context/input separately from lifetime token totals.

Issue #120, opened 2026-05-13
OpenCode session discovery fails on macOS because lsof cwd lookup uses OR semantics without -a


Alexander-He:
## Summary

OpenCode sessions may not be discovered on macOS even when:

- an `opencode` process is actively running
- `~/.local/share/opencode/opencode.db` exists
- `sqlite3` is available
- the session row exists in the OpenCode SQLite database

The root cause appears to be the macOS cwd lookup in `src/collector/opencode.rs`.

## Environment

- abtop: `0.4.3`
- OS: macOS
- OpenCode: `1.14.48`

## What I observed

On my machine:

- `opencode` is running
- `sqlite3` is installed
- `~/.local/share/opencode/opencode.db` exists and contains the current session
- `abtop --once` still shows:

```text
abtop — 0 sessions, 0 mcp servers
```

The current OpenCode session row in the DB has:

- `directory = /Users/alexanderhe`

The running `opencode` process is also in:

- `/Users/alexanderhe`

So discovery should succeed, but it does not.

## Suspected root cause

In `src/collector/opencode.rs`, the macOS cwd lookup uses:

```rust
lsof -p <pid> -d cwd -Fn
```

On macOS, `lsof` selection terms are OR-ed unless `-a` is provided.

That means this command does **not** mean “cwd for this pid”.
It effectively means something closer to:

- entries matching `pid=<pid>` OR
- entries matching `fd=cwd`

As a result, the output can include cwd entries from many unrelated processes.

The current code then takes the first `n...` line:

```rust
stdout
    .lines()
    .find(|l| l.starts_with('n') && l.len() > 1)
    .map(|l| l[1..].to_string())
```

If that first `n...` line belongs to another process, `abtop` gets the wrong cwd, fails to match it against the OpenCode DB session `directory`, and the session is dropped.

## Why fallback did not help

The fallback matcher also checks whether the command line contains the session directory:

```rust
if cmd.contains(session_dir)
```

But when OpenCode is launched simply as:

```text
opencode
```

the command line does not include the cwd, so this fallback also fails.

## Proposed fix

Use `-a` when invoking `lsof` for cwd discovery on macOS:

```bash
lsof -a -p <pid> -d cwd -Fn
```

That makes the filters apply with AND semantics and should return only the cwd entry for the target process.

## Suggested patch

In `get_process_cwd()` for non-Linux targets, change:

```rust
.args(["-p", &pid.to_string(), "-d", "cwd", "-Fn"])
```

to:

[...]

PR #121, opened 2026-05-13
fix(opencode): AND lsof selection terms on macOS for cwd lookup

tbouquet:
Credit @Alexander-He for the diagnosis in #120.

Adds `-a` to the lsof call in `get_process_cwd` so `-p <pid>` and `-d cwd` AND instead of OR. Without it, lsof returns cwd entries from unrelated processes too, and the parser grabs the first `n...` line it sees, which can belong to a different pid. The OpenCode session match against the DB `directory` column then fails and the session gets dropped.

Linux path (`/proc/<pid>/cwd`) is unaffected.

Refs #120

### DIFF
--- a/src/collector/opencode.rs
+++ b/src/collector/opencode.rs
@@ -426,8 +426,10 @@ fn get_process_cwd(pid: u32) -> Option<String> {
 
 #[cfg(not(target_os = "linux"))]
 fn get_process_cwd(pid: u32) -> Option<String> {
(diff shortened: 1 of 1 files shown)


graykode (approved):
Looks good. I verified the macOS lsof behavior locally: without -a, cwd entries from unrelated processes can be returned first, and with -a the lookup is scoped to the target PID's cwd. CI is green, and I don't see any blocking issues.

PR #129, opened 2026-05-25
feat: track active Codex Desktop rollouts

YanxinXue:
## Summary

This improves Codex Desktop session tracking in abtop.

Codex Desktop uses a long-lived `codex app-server` process, and that process may keep rollout JSONL files open after a thread is no longer active. Previously, file-descriptor presence alone could make stale Desktop rollouts appear as active sessions.

This PR handles Codex Desktop separately from regular Codex CLI sessions:

- Detects `codex app-server` processes separately from normal Codex CLI processes.
- Shows only fresh rollout JSONL files whose metadata identifies them as `Codex Desktop`.
- Avoids attributing app-server memory, child processes, and ports to individual Desktop rollout sessions.
- Prevents the kill action from targeting the shared Codex Desktop app-server process.
- Extends the MCP active-rollout freshness threshold to 30 minutes.
- Keeps Codex quota data focused on account-level rate limits instead of model-specific limits.
- Adds tests for Desktop rollout discovery, stale rollout filtering, kill safety, and rate-limit parsing.

## Why

This keeps abtop's session list focused on currently active Codex Desktop threads while avoiding stale entries from old rollouts that the Desktop app-server still has open.

It also keeps destructive session actions conservative: a shared Desktop app-server should not be treated the same way as an individual CLI session process.

## Testing

- `cargo test`
  - 152 passed

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -694,7 +694,7 @@ impl App {
         // Check if we have a pending confirmation for this exact session
         if let Some((idx, ts)) = self.kill_confirm.take() {
             if idx == self.selected && ts.elapsed().as_secs() < 2 {
-                // Confirmed — verify PID still runs expected binary before killing
+                // Confirmed — verify PID still runs a killable agent before killing
                 let pid = session.pid;
                 let verified = std::process::Command::new("ps")
                     .args(["-p", &pid.to_string(), "-o", "command="])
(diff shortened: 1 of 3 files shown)


graykode (approved):
lgtm — security looks clean: no new network/dependency/workflow surface, and the kill path gets stricter for the shared Codex Desktop app-server. I reviewed the Desktop rollout filtering, rate-limit gating, and process ownership changes; local cargo test, cargo clippy -- -D warnings, and cargo build all pass.

PR #130, opened 2026-05-29
fix: improve Codex Desktop rollout discovery

YanxinXue:
## Summary

This PR fixes UI stalls when many Codex Desktop `app-server` processes are running.

Previously, Codex Desktop rollout discovery could call `lsof -F pn -p...` synchronously on the TUI refresh path. On macOS, that command can take tens of seconds when many Desktop threads/processes are open, which makes startup, refresh, panel switching, and keyboard interaction feel frozen.

## Changes

- Move Codex Desktop rollout fd discovery to a background scanner.
- Add a 90-second timeout for the Desktop `lsof` worker.
- Keep the previous successful fd cache when a scan fails or times out.
- Ensure only one Desktop rollout fd scan can run at a time.
- Use cached Desktop fd data through `SharedProcessData.desktop_rollout_fd_map`.
- Avoid changing regular Codex CLI session discovery; `map_pid_to_jsonl()` still follows the existing synchronous CLI path.
- Add filesystem-based recent Desktop rollout discovery so the first screen is not empty while fd scanning is still pending.
- Include recent active Desktop sessions from older date directories, not only today's directory.
- Add an `Unknown` session status for recent Desktop sessions whose rollout file is visible but PID ownership is not yet confirmed.
- Preserve existing active mtime filtering, seen rollout deduplication, and MCP-owned rollout exclusion.

## Why

Codex Desktop can keep many rollout files and process fds around. Blocking the main TUI thread on `lsof` makes the app appear broken or unresponsive. The new path lets the UI render immediately using filesystem-discovered active rollouts, while fd ownership is filled in asynchronously when available.

## Validation

- `cargo test`
- `cargo build`
- `cargo build --release`
- Built test binary:
  `/Users/yanxin/GitHub/abtop/target/distrib/abtop-codex-desktop-fix`

## Risk

Low to moderate.

The change is scoped to Codex Desktop/app-server rollout discovery. Regular Codex CLI, Claude, OpenCode, ports, git stats, and MCP panel behavior are intentionally left mostly unchanged. The main tradeoff is that some Desktop sessions may briefly show `Unknown` until the background fd scan confirms ownership.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -699,7 +699,7 @@ impl App {
             return;
         }
         let session = &self.sessions[self.selected];
-        if session.status == SessionStatus::Done {
+        if matches!(session.status, SessionStatus::Done | SessionStatus::Unknown) {
             return;
         }
 
--- a/src/collector/codex.rs
+++ b/src/collector/codex.rs
@@ -10,6 +10,8 @@ use std::io::{BufRead, BufReader, Read};
 use std::path::{Path, PathBuf};
 #[cfg(all(not(target_os = "linux"), not(target_os = "windows")))]
 use std::process::Command;
+use std::sync::mpsc::{self, Receiver, Sender};
+use std::time::{Duration, Instant};
 
 /// Collector for OpenAI Codex CLI sessions.
 ///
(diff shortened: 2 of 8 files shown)


graykode (approved):
lgtm — security looks clean; this stays within local Codex Desktop rollout discovery and adds no new network, dependency, or workflow surface. I checked the async scanner state flow and Unknown session handling, and the added tests cover the main ownership/fallback cases. Verified with cargo test, cargo clippy --all-targets -- -D warnings, and cargo build.

PR #132, opened 2026-06-02
fix:Chinese i18n

Iamliuxiaozhen:
The official Chinese name for token is '词元' rather than the literal translation of token, '令牌'.


graykode (changes requested):
security looks clean — this is only zh locale string updates, with no network/shell/dependency/workflow changes. I’d hold this until the terminology change is complete: the zh locale still renders `col.tokens` as `Token` and `context.token_rate` as `Token 速率` in `src/locale.rs`, so users will still see the old English token label in the session table and context panel. `cargo test` and `cargo clippy -- -D warnings` pass locally.

PR #134, opened 2026-06-10
fix(windows): host metrics via sysinfo + OpenCode session discovery

XKHoshizora:
## Summary
On Windows 11 abtop showed **"n/a" host metrics** and **0 OpenCode sessions** even with an agent running (reported by an abtop-web-ui tester; reproduces with the TUI and `abtop --json`). This PR fixes both **on Windows only** — every runtime change is `#[cfg(target_os = "windows")]`-gated and the Linux `/proc` path and macOS stub are byte-for-byte unchanged.

## Host metrics (`src/host_info.rs`)
`HostSampler::sample()` returned `None` on every non-Linux platform. `sysinfo` is already a Windows-only dependency, so the sampler now holds a `sysinfo::System` across ticks (CPU usage is a delta between two refreshes — mirrors the Linux prev/delta design): `cpu_pct` = global CPU (0.0 on the first tick, same as Linux), `mem_pct` = used/total, `load1` = 0.0 (Windows has no load average; rendered as N/A).

## OpenCode session discovery (`src/collector/opencode.rs`)
Three Windows-only root causes, found by debugging on a real Windows 11 machine:
1. **`get_process_cwd()` shelled out to `lsof`** (the `not(linux)` branch), which doesn't exist on Windows → cwd matching always failed. New `#[cfg(windows)]` impl reads the cwd via sysinfo (process PEB), refreshing only the queried PID.
2. **Path-format mismatch**: the OpenCode DB stores `directory` with forward slashes (`C:/Users/x/proj` — verified against a real DB) while the live process cwd has backslashes, and NTFS is case-insensitive. Windows-gated `paths_equal()` normalizes separators + case; unix keeps the exact comparison.
3. **Silent empty result when the `sqlite3` CLI is missing** — the common case on Windows. When the DB exists but sqlite3 isn't on PATH, a one-time stderr warning names the DB path and the install command (`winget install SQLite.SQLite`).

Also: `%LOCALAPPDATA%` / `%APPDATA%` are probed as DB-path fallbacks (npm-installed OpenCode keeps the XDG-style `~/.local/share` layout on Windows, so the existing default already resolves), and the project-name fallback goes through `last_path_segment()` so backslash paths don't leak a full path as the project name.

## Windows test suite (pre-existing failures)
`cargo test` had 18 pre-existing failures on Windows — two **fixture** bugs, not product bugs: `claude.rs write_session_file()` built JSON with `format!` (backslash paths → invalid escapes; 14 tests saw zero sessions), now serialized with `serde_json`; `codex.rs set_modified()` opened the file read-only before `File::set_modified()` (`PermissionDenied` on Windows; 4 panics), now opens with write access like `claude.rs set_mtime()` already did.

## Verification
- Windows 11 (MSVC): `cargo build` + `cargo test` green (163/163); `abtop --json` shows real CPU/MEM and the OpenCode session matched by cwd; warning path verified with sqlite3 removed from PATH
- Linux: `cargo build` + `cargo test` green (165 passed + doc-test, 0 failed, verified on real Linux x86_64 before opening this PR); `abtop --json` confirmed unchanged — host metrics from `/proc`, session discovery identical
- macOS: `cargo check --target aarch64-apple-darwin` passes (no code path changes outside `#[cfg(windows)]`)
- README: Windows section + OpenCode notes updated

Thanks for abtop!

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ cargo install abtop
 
 ### Windows
 
-Native support — no WSL required. Uses `sysinfo` for process info and `netstat -ano` for listening ports.
+Native support — no WSL required. Uses `sysinfo` for process info and host CPU/MEM metrics, and `netstat -ano` for listening ports. Windows has no load average, so LOAD is reported as 0. OpenCode session discovery additionally requires the `sqlite3` CLI (`winget install SQLite.SQLite`); without it abtop prints a one-time warning to stderr.
 
 ```powershell
 powershell -c "irm https://github.com/graykode/abtop/releases/latest/download/abtop-installer.ps1 | iex"
@@ -83,7 +83,7 @@ tmux new -s work
 | Subagents         |     ✅      |    ❌     |    ❌    |
 | Memory Status     |     ✅      |    ❌     |    ❌    |
 
-OpenCode support reads the local SQLite database at `~/.local/share/opencode/opencode.db` and requires `sqlite3` in `PATH`.
+OpenCode support reads the local SQLite database at `~/.local/share/opencode/opencode.db` (also the default location on Windows; `%LOCALAPPDATA%\opencode` and `%APPDATA%\opencode` are probed as fallbacks) and requires `sqlite3` in `PATH` (on Windows: `winget install SQLite.SQLite`).
 
 ## Themes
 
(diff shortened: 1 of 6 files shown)


graykode (approved):
lgtm — security is clean: local process/SQLite/sysinfo changes only, with no new network, deps, or workflow surface. I pushed 31b70be to keep Windows LOAD at the documented 0.0 value and refresh the snapshot comment; local build/clippy/tests and refreshed CI are green.

PR #138, opened 2026-06-11
feat: jump to session via terminal adapter registry (cmux/tmux/iTerm2)

### LINKED ISSUE #137: Jump to session beyond tmux (iTerm2, cmux) via terminal adapter registry
**Describe the feature**

Extend the `Enter` "jump to session" so it works outside tmux. Today `Enter` is a no-op unless abtop runs inside tmux. This adds focusing the terminal that runs the selected agent for **iTerm2** and **cmux**, via a small `TerminalJumper` registry (one adapter per backend) so further terminals (WezTerm, Kitty, Ghostty, …) can be added easily.

**Use case**

Many of us run agents in iTerm2 tabs/panes or in cmux workspaces rather than tmux, so the existing jump never fires. With the registry:

- **iTerm2** — match the agent process's controlling tty to an iTerm2 session, focus it via AppleScript.
- **cmux** — read `CMUX_WORKSPACE_ID` from the process env, then `cmux select-workspace --workspace <uuid>`.
- **tmux** — existing behavior preserved.

**Additional context**

- Refactors the inline tmux logic into `src/jump/` (one file per adapter). A three-way `JumpAttempt` (`NotApplicable` / `Jumped` / `Failed`) makes adapters composable — first applicable wins, `NotApplicable` falls through to the next.
- iTerm2's first jump triggers a one-time macOS Automation permission prompt.
- Implementation is ready and verified: 15 new unit tests, live-tested on macOS/iTerm2, and the full CI gate (`clippy -D warnings` / `test` / `build --release`) passes locally on rustc 1.96.0. Happy to open a PR.

Valgard:
## What

`Enter` "jump to session" worked only inside tmux. This refactors the inline tmux logic into a `TerminalJumper` registry under `src/jump/` (one adapter per file) and adds two backends:

- **cmux** — read `CMUX_WORKSPACE_ID` from the agent process env, then `cmux select-workspace --workspace <uuid>`.
- **iTerm2** — match the process's controlling tty to an iTerm2 session and focus it via AppleScript (select pane/tab/window + activate).
- **tmux** — existing behavior preserved (now an adapter; "PID not in any pane" became `NotApplicable` so other backends can try).

## How

`resolve()` walks `jumpers()` in order (cmux → tmux → iTerm2). The three-way `JumpAttempt` (`NotApplicable` / `Jumped` / `Failed`) makes adapters composable — first applicable wins, `NotApplicable` falls through. Backends are mutually exclusive by controlling tty.

## Tests / verification

- 15 new unit tests for the registry loop and the parsing helpers.
- Live-tested on macOS / iTerm2 — focus lands on the correct pane/tab/window.
- Full CI gate green locally on **rustc 1.96.0**: `cargo clippy -- -D warnings`, `cargo test` (180 passed), `cargo build --release`.

## Notes

- iTerm2's first jump triggers a one-time macOS Automation permission prompt; until granted it surfaces as a status message, not a crash.
- Docs updated in `AGENTS.md` (Key Bindings + new "Terminal Jump" section).

Closes #137

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -274,7 +274,7 @@ Tracks child processes that have open ports. When a parent session dies but the 
 | Key | Action |
 |-----|--------|
 | `↑`/`↓` or `k`/`j` | Select session in list |
-| `Enter` | Jump to session terminal (tmux only) |
+| `Enter` | Jump to session terminal (cmux / tmux / iTerm2) |
 | `x` | Kill selected session (SIGKILL) |
 | `X` | Kill all orphan ports |
 | `q` | Quit |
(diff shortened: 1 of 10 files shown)


graykode (approved):
lgtm — security is clean: no new deps/workflows/network calls, and command execution stays on direct argv. I pushed one small fix to keep iTerm2 macOS-only on the registry path and align the footer/help/docs with the expanded terminal-jump behavior.

PR #139, opened 2026-06-14
feat: add Hermes Agent session collector

goelahradhikami-cmyk:
## Summary

Adds a new collector for [Hermes Agent](https://github.com/NousResearch/hermes-agent) — an open-source AI agent framework by Nous Research. Hermes belongs to the same category as Claude Code, Codex CLI, and OpenCode.

## How it works

1. **Process discovery**: Scans running processes for `python -m tui_gateway.slash_worker --session-key <id>` to map PIDs → Hermes session IDs
2. **Session data**: Reads Hermes' SQLite state.db (`~/.hermes/state.db` or `~/AppData/Local/hermes/state.db` on Windows desktop) for session metadata, token counts, model info
3. **PID→session matching**: Connects the two by matching session IDs
4. **Status inference**: Uses process CPU + recent activity to determine Thinking/Executing/Waiting/Done
5. **Context window**: Built-in model→context_size lookup table for 20+ models (configurable)

## Supported features

| Feature | Status |
|---------|--------|
| Session Discovery | ✅ |
| Token Tracking (input/output/cache) | ✅ |
| Status Detection | ✅ |
| Context Window % | ✅ (model lookup table) |
| Child Processes | ✅ |
| Git Status | ✅ (via existing MultiCollector git sweep) |
| Rate Limit | ❌ (Hermes doesn't expose this) |

## Usage

Hermes sessions appear automatically with agent_cli `"hermes"`. To hide them:
```toml
hidden_agents = ["hermes"]
```

## Cross-platform

Tested on Windows. Works on Linux and macOS via the same sqlite3 CLI approach used by OpenCodeCollector.

Closes #...

### DIFF
--- a/src/collector/hermes.rs
+++ b/src/collector/hermes.rs
@@ -6,8 +6,11 @@ use std::fs;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
-/// Maximum sessions to fetch from the DB per query.
-const MAX_SESSIONS: u32 = 20;
+/// Maximum sessions to fetch from the DB per query. With 140+ active Hermes
+/// sessions, a simple `LIMIT` would miss older sessions that still have
+/// running workers, so we use a two-phase approach: collect running PIDs
+/// first, then query only those session IDs.
+const MAX_SESSIONS: u32 = 100;
 
 /// Model -> context window size (tokens) lookup.
 /// Hermes doesn't store context_window in the DB, so we maintain a table.
@@ -73,8 +76,8 @@ fn expand_home(path_str: &str) -> PathBuf {
 /// so we don't fork a sqlite3 process every 2s.
 pub struct HermesCollector {
     db_path: PathBuf,
-    /// Whether sqlite3 CLI is available (checked once).
-    sqlite3_available: Option<bool>,
+    /// Whether Python (with sqlite3 module) is available (checked once).
+    python_available: Option<bool>,
     /// Cached DB rows from the last slow-tick query.
     cached_db_sessions: Vec<DbSession>,
 }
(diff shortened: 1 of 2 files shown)


graykode (changes requested):
held off approving — security is not clean. The new Hermes collector builds Python source with formatted DB paths/session IDs in  and ; values from , process args, or the DB can contain quotes/newlines and break out of those raw Python strings, so this needs to pass data via argv/stdin/JSON or use a non-codegen query path. It also stores Hermes initial prompts, first assistant text, and tool-call args without the same redaction/sanitization used by other collectors (, , ), which can leak transcript secrets through the TUI/JSON snapshot. The temp SQL file is predictable and written outside the project (), and  currently fails.

graykode:
Clarifying my request-changes review above with the exact references:

Security is not clean. The new Hermes collector builds Python source with formatted DB paths/session IDs in `src/collector/hermes.rs:192` and `src/collector/hermes.rs:695`; values from `HERMES_HOME`, process args, or the DB can contain quotes/newlines and break out of those raw Python strings, so this needs to pass data via argv/stdin/JSON or use a non-codegen query path.

It also stores Hermes initial prompts, first assistant text, and tool-call args without the same redaction/sanitization used by other collectors (`src/collector/hermes.rs:758`, `src/collector/hermes.rs:769`, `src/collector/hermes.rs:781`), which can leak transcript secrets through the TUI/JSON snapshot.

The temp SQL file is predictable and written outside the project (`src/collector/hermes.rs:183`), and `cargo clippy --all-targets -- -D warnings` currently fails.

graykode:
Thanks for the substantial work on the Hermes collector. After a closer security review, I’m not comfortable merging this implementation as-is: it adds new local Python execution, SQLite parsing, temp-file, and transcript-display surfaces that would need more hardening and maintenance.

For now, we also don’t have a strong enough reason to introduce Hermes support into abtop, so I’m going to close this PR rather than ask you to keep iterating on a large integration we may not adopt. If Hermes support becomes a project priority later, we can revisit it with a smaller design that avoids dynamic code generation and follows the existing collector privacy/redaction model from the start.

PR #157, opened 2026-07-17
fix: keep parent active for async subagents

### LINKED ISSUE #156: Claude parent shows Waiting while async subagents are still working
## Bug description

When Claude Code is waiting for asynchronously launched `Agent` subagents, abtop reports the parent session as `Waiting` even though the subagents are still actively working and updating their transcripts.

This makes the top-level status misleading when monitoring multiple long-running Claude Code sessions: `◌ Wait` normally means the session needs user input, but in this case no input is needed and work is still in progress.

## Steps to reproduce

1. Start a Claude Code session.
2. Ask Claude to launch one or more `Agent` subagents.
3. With current Claude Code, observe that the `Agent` tool receives an immediate `tool_result` saying that an async/background agent was launched, including an `agentId` and output file.
4. Let the parent emit a text-only assistant message while it waits for the subagent results.
5. Open abtop (optionally enable the subagent tree with `T`).

## Expected behavior

The parent session should remain active, preferably `Executing`, while any of its subagents is still `working`.

## Actual behavior

The parent session changes to `Waiting` while its subagent transcripts continue to update. The subagent tree may simultaneously show working subagents under a parent marked `Wait`.

## Reproduction evidence

I reproduced this from a real session with two anonymized subagents:

- The first `Agent` tool received its async result about 176 ms after launch, but its subagent continued writing for another 14m 59s.
- The second received its async result about 61 ms after launch, but continued writing for another 13m 40s.
- After the parent's later text-only assistant message, the two subagents remained active for about 14m 02s and 13m 08s respectively, during which the parent could be classified as `Waiting`.

A minimal regression fixture using this transcript sequence also reproduces it:

```text
user prompt
assistant Agent tool_use
user tool_result: async agent launched, running in background
assistant text-only message: waiting for research result
subagent JSONL still recently updated
```

The expected parent status was `Executing`, but the result was:

```text
left: Waiting
right: Executing
```

## Root cause analysis

The Claude collector currently determines parent status using only:

- active descendant CPU,
- a non-empty `current_task` (pending tool), or
- a trailing real user message (`last_user_ts_ms`).

Relevant code:

- [Parent status calculation](https://github.com/graykode/abtop/blob/a3328db20fe189887986e0147080954bc07178a1/src/collector/claude.rs#L539-L571)
- [`current_task` is cleared on every assistant entry](https://github.com/graykode/abtop/blob/a3328db20fe189887986e0147080954bc07178a1/src/collector/claude.rs#L1425-L1430)
- [Subagents are collected only after parent status has already been calculated](https://github.com/graykode/abtop/blob/a3328db20fe189887986e0147080954bc07178a1/src/collector/claude.rs#L619-L628)
- [`collect_subagents` already derives `working`/`done`](https://github.com/graykode/abtop/blob/a3328db20fe189887986e0147080954bc07178a1/src/collector/claude.rs#L680-L768)

For an async `Agent` call, Claude Code writes the tool result immediately. A subsequent text-only assistant entry clears `current_task`, and the subagent may not appear as a CPU-active descendant process. The status therefore falls through to `Waiting` even though `collect_subagents()` can see a recently active subagent.

PRs #84 and #101 fixed a closely related case where a normal `tool_use` remains pending, and their existing regression test passes. However, async `Agent` calls return their `tool_result` immediately, so that pending-tool signal cannot represent the remaining subagent lifetime.

## Possible fix

As a minimal fix, collect subagents before calculating parent status and classify the parent as `Executing` when any subagent is `working`.

The existing 30-second transcript-mtime heuristic may still flicker during a long subagent model generation. A more robust implementation could track the async lifecycle by linking the `toolUseId` in `*.meta.json` with the `agentId` returned by the parent `Agent` tool result.

## Environment

- abtop: `0.5.3`
- abtop commit: `a3328db20fe189887986e0147080954bc07178a1` (current `main` and `v0.5.3`)
- Claude Code: `2.1.212`
- OS: Ubuntu 24.04, Linux 6.8.0-117-generic x86_64

kouyichi:
## Summary

- collect Claude subagents before calculating the parent session status
- keep the parent session `Executing` while any async subagent is still `working`
- add a regression test covering an immediate async `Agent` tool result followed by a text-only parent response

## Problem

Async Claude `Agent` calls return a `tool_result` immediately even though the subagent continues in the background. Once the parent writes a text-only response, `current_task` is cleared and abtop falls through to `Waiting`, despite already discovering a working subagent for the session.

## Test plan

- `cargo clippy -- -D warnings`
- `cargo test`
- `cargo build --release`

Closes #156

### DIFF
--- a/src/collector/claude.rs
+++ b/src/collector/claude.rs
@@ -536,13 +536,28 @@ impl ClaudeCollector {
             return None;
         }
 
+        // Derive the project directory from the transcript path (handles worktree sessions),
+        // falling back to the encoded cwd.
+        let project_dir = transcript_path
+            .as_ref()
+            .and_then(|tp| tp.parent().map(|p| p.to_path_buf()))
+            .unwrap_or_else(|| config.projects_dir.join(encode_cwd_path(&sf.cwd)));
+
+        // Collect subagents before deriving the parent status so asynchronous
+        // Agent work keeps the parent active after its tool_result has returned.
+        let subagents_dir = project_dir.join(&sf.session_id).join("subagents");
+        let subagents = Self::collect_subagents(&subagents_dir);
+        let has_working_subagent = subagents.iter().any(|agent| agent.status == "working");
+
         // Status is best-effort. Signals we trust:
         //   1. Active descendant CPU → tool is running.
         //   2. current_task non-empty → latest assistant turn left a
         //      tool_use unanswered. Catches I/O-bound tools (Read, Edit)
         //      whose descendants stay under 5% CPU, so the CPU heuristic
         //      alone would flicker to Waiting while the tool runs.
-        //   3. last_user_ts_ms > 0 → trailing transcript line is a real
+        //   3. A working subagent → an async Agent call is still running even
+        //      though its tool_result has already returned to the parent.
+        //   4. last_user_ts_ms > 0 → trailing transcript line is a real
         //      user prompt with no assistant reply yet, so the model is
         //      generating. tool_result wrappers are skipped at the
         //      parser level so this only fires for actual prompts.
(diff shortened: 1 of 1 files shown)


graykode (approved):
Reviewed the status transition, subagent activity heuristic, filesystem behavior, and regression coverage. The change is focused, preserves the existing collection path, and correctly keeps the parent session Executing while an asynchronous subagent remains active. The full test suite, clippy, and release build pass locally. Approved.

PR #161, opened 2026-07-20
fix(codex): track code-mode status lifecycle

### LINKED ISSUE #160: Codex 0.144.6 status detection misses Code Mode lifecycle events
## Environment

- abtop 0.5.3
- Codex CLI 0.144.6
- Linux

## Problem

Codex session status frequently flips to `Waiting` while a turn is still active, and some waits are shown as `Executing`.

Current Codex Code Mode rollouts use `response_item.custom_tool_call` and `custom_tool_call_output`, while the collector only tracks `function_call` and `function_call_output`. The rollout also emits intermediate `agent_message` events before continuing to reason or invoke another tool, but the collector currently treats every `agent_message` as the end of the active turn.

This causes several observable errors:

- Code Mode tool calls are absent from the timeline and can be reported as `Waiting` while executing.
- A progress `agent_message` closes the Thinking window even though the same turn continues.
- After a tool result, continued model reasoning can be reported as `Waiting`.
- A pending `request_user_input` call is reported as `Executing` instead of `Waiting`.
- Busy persistent children such as `codex-code-mode-host` can override explicit rollout state and report `Executing` after a turn has completed.
- Automatic/subagent turns that start with `task_started` but have no `user_message` never enter Thinking.

## Expected behavior

Explicit rollout lifecycle events should take precedence over CPU heuristics:

1. A completed turn or an explicit user-input request is `Waiting`.
2. An open ordinary tool call is `Executing`.
3. An active turn with no open tool is `Thinking`.
4. Descendant CPU activity is only a fallback when the rollout has no explicit active state.

Both standard and custom tool-call event forms should participate in the same lifecycle and timeline tracking.

## Reproduction

1. Start abtop 0.5.3 and Codex CLI 0.144.6 with Code Mode enabled.
2. Ask Codex to inspect a repository and run several commands.
3. Observe abtop after an intermediate progress message and between a tool result and the next action.
4. Trigger `request_user_input` and observe that it is shown as `Executing`.

The status differs from the actual rollout state in each case above.

kouyichi:
## Summary

- recognize `custom_tool_call` and `custom_tool_call_output` events emitted by current Codex Code Mode
- keep turns active from `task_started`/`user_message` through `task_complete`, including automatic turns with no user message
- transition back to Thinking after the final tool result while a turn is still active
- report explicit user-input requests and completed turns as Waiting, with descendant CPU used only as a fallback
- preserve standard tool-call and asynchronous `exec_command`/`write_stdin` tracking
- derive current-task text from the final session status so Done and Unknown sessions retain consistent labels

## Root cause

The collector treated every `agent_message` as the end of a turn and only parsed the older `function_call` event shape. Codex CLI 0.144.6 emits intermediate progress messages and routes Code Mode tools through custom tool-call events, so the parsed state frequently disagreed with the live turn.

## Testing

- `cargo clippy --all-targets -- -D warnings`
- `cargo test` (207 unit tests and 1 doc test)
- manual `abtop --json` validation against active Codex CLI 0.144.6 sessions covering active tools, post-tool reasoning, completed turns, and `request_user_input`

Closes #160

### DIFF
--- a/src/collector/codex.rs
+++ b/src/collector/codex.rs
@@ -567,10 +567,14 @@ impl CodexCollector {
                 && process_ctx.pid.is_some_and(|p| {
                     process::has_active_descendant(p, children_map, process_info, 5.0)
                 });
-            if has_active_child || result.pending_since_ms > 0 {
+            if result.task_complete || result.waiting_for_user {
+                SessionStatus::Waiting
+            } else if result.pending_since_ms > 0 {
                 SessionStatus::Executing
             } else if result.model_generating {
                 SessionStatus::Thinking
+            } else if has_active_child {
+                SessionStatus::Executing
             } else {
                 SessionStatus::Waiting
             }
@@ -579,14 +583,14 @@ impl CodexCollector {
         // Current task from last tool use
         // For exec (one-shot) sessions, task_complete means truly finished.
         // For interactive sessions, task_complete fires after every turn — ignore it.
-        let current_tasks = if !result.current_task.is_empty() {
-            vec![result.current_task]
-        } else if matches!(status, SessionStatus::Unknown) {
+        let current_tasks = if matches!(status, SessionStatus::Unknown) {
             vec!["unknown".to_string()]
-        } else if !pid_alive || (process_ctx.is_exec && result.task_complete) {
+        } else if matches!(status, SessionStatus::Done) {
             vec!["finished".to_string()]
         } else if matches!(status, SessionStatus::Waiting) {
             vec!["waiting for input".to_string()]
+        } else if !result.current_task.is_empty() {
+            vec![result.current_task]
         } else {
             vec!["thinking...".to_string()]
         };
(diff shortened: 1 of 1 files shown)


graykode (changes requested):
Thanks for the comprehensive lifecycle work. The status ordering and custom tool-call tracking are generally well structured, and the full test suite, clippy, and release build pass locally. However, I found two blocking issues in the current Code Mode path.

1. The user-input wait detection does not match the event shape emitted by Code Mode. tool_waits_for_user checks the recorded tool name for request_user_input or AskUserQuestion, but current Code Mode records this as a custom_tool_call whose name is exec and whose input contains tools.request_user_input(...). call_names therefore stores exec, waiting_for_user remains false, and the open custom call is reported as Executing instead of Waiting. In sampled current rollouts, the user-input requests were custom-wrapped and there were no standalone function_call events for request_user_input. The new regression test uses the standalone function_call shape, so it does not reproduce the actual failure. Please add a fixture using custom_tool_call name exec with tools.request_user_input in input and make that path resolve to Waiting.

2. The new custom-tool branch copies the entire raw JavaScript input into ToolCall.arg after only secret-prefix redaction and a 120-character truncation. sanitize_tool_arg does not call sanitize_terminal_text and does not extract a bounded safe argument. This value reaches current_task, the TUI timeline, and JSON snapshots. Ordinary Code Mode inputs are multiline, and attacker-controlled ANSI/control or bidi characters can pass through as well; the raw source can also contain request bodies or other content that should not be treated as a short argument preview. Please sanitize terminal controls and bidi marks, and preferably extract only an allowlisted path or command preview. If a safe preview cannot be derived, display the tool name without the raw input. Add regression coverage for multiline input, terminal control characters, bidi marks, and sensitive inline content.

The lifecycle changes outside these two paths look sound. My requested changes are limited to matching the real custom-wrapped user-input event and keeping raw Code Mode source out of display and snapshot fields.

graykode (approved):
The custom-tool payload disclosure is fixed by 559dc1c: opaque inputs no longer reach task descriptions or tool argument previews, with regression coverage. The latest head is exactly that reviewed change merged with main, and CI passes. The previously noted custom-wrapped user-input detection remains a functional limitation (Executing instead of Waiting); I am treating it as a non-blocking follow-up rather than a security blocker. Approving this revision for merge.

